PR #2, opened 2019-06-14
add Dockerfile.rhel7 for OpenShift build

zshi-redhat:


zshi-redhat:
@ahalim-intel @przemeklal PTAL

przemeklal:
lgtm

ahalim-intel:
@przemeklal I have one comment, not about this PR but related. 
I see, existing Dockerfile is under `build/` directory. Maybe we should consolidate all docker files in same location.

przemeklal:
in most of our projects we keep in the top level so I moved it there as well, please see PR #3

PR #6, opened 2019-06-24
add build binary script

zshi-redhat:
This also defaults `make` to build binary instead of
image; To build container image, run image build
explicitly `make image`.


zshi-redhat:
@przemeklal @ahalim-intel PTAL

przemeklal:
thanks @zshi-redhat, looks good to me

PR #9, opened 2019-07-01
support third-party certificate

zshi-redhat:


scripts/webhook-patch-ca-bundle.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# Original script found at: https://github.com/morvencao/kube-mutating-webhook-tutorial/blob/master/deployment/webhook-patch-ca-bundle.sh
+
+ROOT=$(cd $(dirname $0)/../../; pwd)
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+export CA_BUNDLE=$(kubectl get configmap -n kube-system extension-apiserver-authentication -o=jsonpath='{.data.client-ca-file}' | base64 | tr -d '\n')
```
przemeklal: minor comment: `base64 -w=0` disables line wrapping, so that you won't have to run `tr -d '\n'`

zshi-redhat:
> /lgtm (one very, very small comment)
> 
> Just thinking - would it be suitable to use something like [this](https://kubernetes.github.io/ingress-nginx/examples/tls-termination/) for the TLS termination or use another similar mechanism e.g. provided by OpenShift? This would allow us to get rid of handling TLS in the application, generating certs/key, signing them, etc.

Yes, good one!
OpenShift uses [service-ca-operator](https://github.com/openshift/service-ca-operator) to issue signed serving certificate/key pair to services via secret and update cabundle of configMap. The secret will be updated when it gets close to expiration automatically. With [secret](https://github.com/intel/network-resources-injector/pull/9/files#diff-a389e926020e75553ce0f76e72a67693R62) attached in `server.yaml` file, the webhook server shall have the valid certificate.

The templates(service.yaml, webhook.yaml, auth.yaml, server.yaml) under `deployments` folder in this PR will be rendered by OpenShift SR-IOV Network Operator (a component that manages e2e sriov configuration including sriov admission controller, sriov device plugin, sriov cni, sriov net-attach-def crd etc). This allows tls cert be done outside application.

scripts/webhook-patch-ca-bundle.sh
zshi-redhat: Done

przemeklal:
+1 and thanks for explaining!

PR #13, opened 2019-07-10
fix missing return at the end of patchEmptyResources

zshi-redhat:


zshi-redhat:
> Apologies for this. It looks like I forgot to add this in my PR for downwardAPI. I had this in my local repo, just missed pushing that last patch with this update.
> 
> Thanks for taking care of this Zenghui.

Hi Victor, no worries, I think this is because the downward API PR was initially tested fine, but merged after the fix of `patchEmptyResources` PR, so we were not able to capture this failure.

przemeklal:
Thanks for the fix! /lgtm

PR #14, opened 2019-07-12
restart http server upon cert file update

zshi-redhat:
When third-party certificate files are injected via secret volume
and be automatically replaced when the secret gets close to
expiration, Webhook server will capture this update and restart
http server.

This will not affect self-signed certificate created via initContainer.

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -69,6 +69,9 @@ func main() {
 	}
 	defer watcher.Close()
 
+	certUpdated := false
+	keyUpdated := false
+
 	for {
 		watcher.Add(*cert)
 		watcher.Add(*key)
@@ -83,8 +86,18 @@ func main() {
 				fsnotify.Write | fsnotify.Chmod
 			if (event.Op & mask) != 0 {
 				glog.Infof("modified file: %v", event.Name)
-				if err := keyPair.Reload(); err != nil {
-					glog.Fatalf("Failed to reload certificate: %v", err)
+				if event.Name == *cert {
+					certUpdated = true
+				}
+				if event.Name == *key {
+					keyUpdated = true
+				}
+				if keyUpdated && certUpdated {
+					if err := keyPair.Reload(); err != nil {
+						glog.Fatalf("Failed to reload certificate: %v", err)
+					}
+					certUpdated = false
+					keyUpdated = false
 				}
 			}
 		case err, ok := <-watcher.Errors:
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -79,12 +79,11 @@ func main() {
 				continue
 			}
 			glog.Infof("watcher event: %v", event)
-			if event.Op & fsnotify.Create == fsnotify.Create ||
-				event.Op & fsnotify.Rename == fsnotify.Rename ||
-				event.Op & fsnotify.Remove == fsnotify.Remove ||
-				event.Op & fsnotify.Write == fsnotify.Write {
+			mask := fsnotify.Create | fsnotify.Rename | fsnotify.Remove |
+				fsnotify.Write | fsnotify.Chmod
+			if (event.Op & mask) != 0 {
 				glog.Infof("modified file: %v", event.Name)
-				if err := keyPair.MaybeReload(); err != nil {
+				if err := keyPair.Reload(); err != nil {
 					glog.Fatalf("Failed to reload certificate: %v", err)
 				}
 			}
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -15,11 +15,13 @@
 package main
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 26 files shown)


cmd/webhook/main.go
```diff
@@ -37,11 +42,45 @@ func main() {
...
+		}
+		newHashVal := hex.EncodeToString(hasher.Sum(nil))
+		if oldHashVal != "" && newHashVal != oldHashVal {
+			glog.Info("get cert file update")
+			if err := httpServer.Shutdown(context.Background()); err != nil {
+				glog.Fatalf("http server shutdown: %v", err)
+			}
+		}
+		oldHashVal = newHashVal
+		time.Sleep(1 * time.Second)
```
przemeklal: do you think that we could watch for the changes in the filesystem (inotify) or watch Kubernetes secret resource instead of recalculating hash every second?

zshi-redhat: +1, agree, I'll see if it could monitor the change instead of doing the cal every second.
(Monitor kubernetes secret may require passing additional info such as secret name into webhook server)

zshi-redhat: @przemeklal ,  Just added `fsnotify` to watch the file change, PTAL

cmd/webhook/main.go
```diff
@@ -33,15 +35,64 @@ func main() {
...
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				continue
+			}
+			glog.Infof("watcher event: %v", event)
+			if event.Op & fsnotify.Create == fsnotify.Create ||
+				event.Op & fsnotify.Rename == fsnotify.Rename ||
+				event.Op & fsnotify.Remove == fsnotify.Remove ||
+				event.Op & fsnotify.Write == fsnotify.Write {
```
przemeklal: question - why do we watch for Remove and Rename? and not for Chmod (the only Op that's left out here)
it could be a bit simpler if we handled any type of operation (including chmod) or only the ones that we need to

But if we need precisely this set of Ops than that's fine.

Also, nit comment on the readability/simplicity - Op is uint32 so maybe we could simplify it a bit and do something like this:
```
mask := fsnotify.Create | fsnotify.Rename | fsnotify.Remove | fsnotify.Write
if (event.Op & mask) != 0 {
    ...
}
```

pkg/webhook/tlsutils.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type tlsKeypairReloader struct {
+	certMutex sync.RWMutex
+	cert      *tls.Certificate
+	certPath  string
+	keyPath   string
+}
+
+func (keyPair *tlsKeypairReloader) MaybeReload() error {
```
przemeklal: why Maybe? :wink:

zshi-redhat: This was to say it reloads certificate files whose content may not be changed.
I'll change it to `Reload()` :-)

cmd/webhook/main.go
zshi-redhat: I tried to capture all the possible events during testing, and found that depending on different operations(rename, move, delete, create etc), it triggers various event, so I'd like to capture all of them and reload the certificate(reloading is cheap, it doesn't require much cycles);
I will also add chmod and use your simplified code:
```
mask := fsnotify.Create | fsnotify.Rename | fsnotify.Remove | fsnotify.Write | fsnotify.Chmod
if (event.Op & mask) != 0 {
    ...
}
```

przemeklal:
Thanks for the update.

I ran it on my cluster and can confirm it works, server reloads after secret update.

Last question - I found that it reloads twice tho - first time it detects the cert file update, and then reloads again after receiving key file update event from the watcher. Should we reload only once when both files have changed to avoid possible race between watcher and the actual file update? If we reload before the key has been updated, we could hit a key and cert mismatch. I noticed that there's usually a small delay (~10-15s) before the watcher generates event, so most likely this isn't going to occur in the reality, but one can never be too confident :)

zshi-redhat:
> Thanks for the update.
> 
> I ran it on my cluster and can confirm it works, server reloads after secret update.
> 
> Last question - I found that it reloads twice tho - first time it detects the cert file update, and then reloads again after receiving key file update event from the watcher. Should we reload only once when both files have changed to avoid possible race between watcher and the actual file update? If we reload before the key has been updated, we could hit a key and cert mismatch. I noticed that there's usually a small delay (~10-15s) before the watcher generates event, so most likely this isn't going to occur in the reality, but one can never be too confident :)


This a good one! 
I think this is caused by two files (key, cert) are updated separately, not at the same time. 
Just updated another version to only reload tls when two files are updated. This assumes that key and cert will always be updated together, not just one.

przemeklal:
Thank @zshi-redhat, nicely done, lgtm

@ahalim-intel I think we can go ahead and merge this PR

ahalim-intel:
Thanks @zshi-redhat, nice work :)
Thanks @przemeklal for your great review and help improving it.
I'll merge this now.

PR #15, opened 2019-11-08
Update multus-cni from v3.2 to ecc474a

vpickard:
Update package multus-cni from v3.2 to current tip of master.

Multus added support for multiple IPs/interface in this
PR: https://github.com/intel/multus-cni/pull/387

The latest available tag is v3.3, which does not include
the above PR.

Signed-off-by: vpickard <vpickard@redhat.com>


vpickard:
https://github.com/intel/network-resources-injector/issues/16

zshi-redhat:
@dougbtv @s1061123 do you think we could have a tag in Multus to include above change (support for multiple IPs/interface PR: intel/multus-cni#387)

vpickard:
> @dougbtv @s1061123 do you think we could have a tag in Multus to include above change (support for multiple IPs/interface PR: [intel/multus-cni#387](https://github.com/intel/multus-cni/pull/387))

I had a chat with @dougbtv regarding adding a 3.3.1 tag before pushing this PR. Neither of us thought it was really necessary, but, if @zshi-redhat  and @ahalim-intel think that is a better approach, I can adjust the PR to use a version tag (once available).

dougbtv:
We'll tag a new version during this Thursday's maintainers meeting for Multus assuming there's consensus. But, for what it's worth... IMO there's no penalty for tracking a commitish in depedency management in this project. Tracking using semver is most useful for projects that you don't want to have to track in detail (to express "I need this version or better, and there's a contract for API compatibility"). Whereas, Multus is a project that you track in detail included as a dependency in this project, and outside of the NPWG specifications, it doesn't offer API promises for use as a library, so you may expect that you have to adjust according to the changes regardless of tagged version.

But, regardless, we'll tag a new version that includes the features you're looking for this Thursday -- November 14th 2019.

PR #21, opened 2020-03-11
Add NodeSelector support

pperiyasamy:
This commit would parse the net-attach-def for any node selector label
present (example: k8s.v1.cni.cncf.io/nodeSelector: environment=ovs)
and inject into pod spec under nodeSelector map.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -569,7 +569,13 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
 			/* parse the net-attach-def annotations for node selector label and add it to the desiredNsMap */
 			if ns, exists := networkAttachmentDefinition.ObjectMeta.Annotations[nodeSelectorKey]; exists {
 				nsNameValue := strings.Split(ns, "=")
-				if len(nsNameValue) == 2 {
+				nsNameValueLen := len(nsNameValue)
+				if nsNameValueLen > 2 {
+					errString := fmt.Sprintf("node selector in net-attach-def %s has more than one label", n.Name)
+					glog.Error(errString)
+					http.Error(w, errString, http.StatusBadRequest)
+					return
+				} else if nsNameValueLen == 2 {
 					desiredNsMap[strings.TrimSpace(nsNameValue[0])] = strings.TrimSpace(nsNameValue[1])
 				} else {
 					desiredNsMap[strings.TrimSpace(ns)] = ""
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -54,6 +54,7 @@ type hugepageResourceData struct {
 
 const (
 	networksAnnotationKey = "k8s.v1.cni.cncf.io/networks"
+	nodeSelectorKey       = "k8s.v1.cni.cncf.io/nodeSelector"
 )
 
 var (
@@ -481,6 +482,26 @@ func createEnvPatch(patch []jsonPatchOperation, container *corev1.Container,
 	return patch
 }
 
+func updateNodeSelector(existing map[string]string, desired map[string]string, patch []jsonPatchOperation) []jsonPatchOperation {
+	newExistingMap := make(map[string]string)
+	if existing != nil {
+		for k, v := range existing {
+			newExistingMap[k] = v
+		}
+	}
+	if desired != nil {
+		for k, v := range desired {
+			newExistingMap[k] = v
+		}
+	}
+	patch = append(patch, jsonPatchOperation{
+		Operation: "add",
+		Path:      "/spec/nodeSelector",
+		Value:     newExistingMap,
+	})
+	return patch
+}
+
 // MutateHandler handles AdmissionReview requests and sends responses back to the K8s API server
 func MutateHandler(w http.ResponseWriter, req *http.Request) {
 	glog.Infof("Received mutation request")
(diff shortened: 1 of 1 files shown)


JanScheurich:
This PR is an attempt to simplify the scheduling of pods referring to a NAD which is not available on all nodes of a K8s cluster, for example a MacVlan NAD whose parent interface exists only on some nodes.

The canonical way to solve the scheduling issue today is for the pod spec to contain an appropriate node selector label along with the network annotation. The right label to use must somehow be documented for the NAD.

The idea in this PR is to hide the need for the extra node selector label from the application and instead specify the label in the NAD and let the Network Resources Injector webhook automatically augment the pod's node selector section, in the same spirit as it does this today for network resources like PCI devices.

The creator of the NAD still has to pick a label and apply it to the the appropriate nodes. Further automation of that process would be left to future network configuration operators.

pperiyasamy:
@zshi-redhat et al: Please take a look and provide your feedback.

pkg/webhook/webhook.go
```diff
@@ -571,6 +608,9 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
 				}
 			}
 
+			/* patch the pod with additional node selector labels */
+			patch = updateNodeSelector(existingNsMap, desiredNsMap, patch)
```
zshi-redhat, line 612: nit: move patch as first argument in updateNodeSelector? so that it's consistent with other patch functions.

pkg/webhook/webhook.go
```diff
@@ -486,6 +513,16 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
 					glog.Infof("network '%s/%s' doesn't use custom resources, skipping...", n.Namespace, n.Name)
 				}
 			}
+
+			/* parse the net-attach-def annotations for node selector labels and add it to the addedNsMap */
+			if ns, exists := networkAttachmentDefinition.ObjectMeta.Annotations[nodeSelectorKey]; exists {
+				nsNameValue := strings.Split(ns, "=")
+				if len(nsNameValue) == 2 {
```
zshi-redhat, line 520: Shall we automatically append the label value if it's define with key only? for example:
In net-attach-def, the annotation is `k8s.v1.cni.cncf.io/nodeSelector: environment`, then NRI update it to `k8s.v1.cni.cncf.io/nodeSelector: environment=""`

pkg/webhook/webhook.go
```diff
@@ -429,6 +430,26 @@ func createVolPatch(patch []jsonPatchOperation, hugepageResourceList []hugepageR
...
+		for k, v := range existing {
+			newExistingMap[k] = v
+		}
+	}
+	if desired != nil {
+		for k, v := range desired {
+			newExistingMap[k] = v
+		}
+	}
+	patch = append(patch, jsonPatchOperation{
```
zshi-redhat, line 445: I assume we don't need to patch nodeSelector path when pod manifest doesn't have nodeSelector defined?  Like how the `patchEmptyResources` function does.

pperiyasamy: correct, now updated the code not to add nodeSelector path when labels map is empty.

pkg/webhook/webhook.go
pperiyasamy: Good find! done.

pkg/webhook/webhook.go
pperiyasamy: done

zshi-redhat (approved):

zshi-redhat:
@martinkennelly PTAL

pkg/webhook/webhook.go
```diff
@@ -486,6 +513,16 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
 					glog.Infof("network '%s/%s' doesn't use custom resources, skipping...", n.Namespace, n.Name)
 				}
 			}
+
+			/* parse the net-attach-def annotations for node selector label and add it to the desiredNsMap */
+			if ns, exists := networkAttachmentDefinition.ObjectMeta.Annotations[nodeSelectorKey]; exists {
+				nsNameValue := strings.Split(ns, "=")
```
martinkennelly, line 519: is it possible to have multiple key-value pairs defined for annotation: k8s.v1.cni.cncf.io/nodeSelector?
e.g. k8s.v1.cni.cncf.io/nodeSelector: disk=ssd,cpu=xeon
I do not have experience with this annotation.
Where is the documentation describing this annotation is what I am trying to get at? I cannot find it.

If it is limited to one key=value pair, then we should be opinionated if len(nsNameValue > 2) and return a bad request error.

pperiyasamy: @martinkennelly Thanks, I too thought about this. so far we've seen a need for having single node selector label for the NAD. for example, select a node in which master interface is present for macvlan cni type, select a node in which ovs bridge is present for ovs-cni type. I think we would go with a check to limit with one key=value pair and throw a bad request error if its more than 1. WDYT ?
@JanScheurich Please let us know your thoughts.

martinkennelly: yeah. I think a limit check is good to prevent weird errors if we aren't supporting multiple key/value pairs in this patch.

pperiyasamy: done.

martinkennelly (approved):
Great patch. Thank you!

PR #23, opened 2020-06-30
Update for k8s 1.18 version

pperiyasamy:
1. Add signers for the init container because signers are needed for CSRs. 
This is mandatory for k8s 1.18 version as per below [PR](https://github.com/kubernetes/kubernetes/pull/86933).

2. Add Response.PatchType into AdmissionReview as it would be needed
for some versions of k8s 1.18 deployment


przemeklal:
low risk change, LGTM

Issue #26, opened 2020-07-13
network resources injector cannot handle pods from deployment properly


zshi-redhat:
network resources injector inspects the net-attach-def and injects the sriov resource request to pod resource request/limits. It works perfectly with a pod, but has issues dealing with a deployment pod in a non-default namespace.

The workflow of network resources injector is:
1) watch the creation/update of *pod*
2) receive AdmissionReview request
3) deserialize pod object from AdmissionReview request
4) set pod namespace to 'default', If pod object doesn't contain valid namespace (e.g. empty namespace for a deployment pod)
5) call API to retrieve net-attach-def with above namespace + net-attach-def name (from pod object annotation)
6) If resource is found in net-attach-def, inject the resource name in pod object. else do nothing.

The problem for a deployment pod happens in step 5),  the deserialized deployment pod object contains an empty namespace field, so it is set to default namespace, which in turn cause the actual namespace of deployment not be respected. So when network-resources-injector tries to get the net-attach-def with default namespace, it fails.

It looks like we have two possible ways to fix this:
1) find a way to get pod namespace from deployment
2) watch the creation/update of deployment object and mutate the deployment spec (this implies that we should ignore the pod creation/update event that's from the same deployment)

ahalim-intel:
cc @przemeklal 
PTAL

pperiyasamy:
@zshi-redhat @przemeklal Let me have a look at this issue and propose a fix (with 1st approach).

JanScheurich:
Is it a bug or a feature of K8s that a pod of a deployment does not inherit the namespace of its deployment?

pperiyasamy:
> Is it a bug or a feature of K8s that a pod of a deployment does not inherit the namespace of its deployment?

@JanScheurich AFAIK, the pod replica doesn't have a namespace when its admission review request is received on the webhook, so NRI must rely on the namespace set at the deployment level so that net-attach-def object can be queried from the appropriate namespace. I'm already working on a patch for it. 
But I can see pod replica has namespace set once it is deployed on the node.

JanScheurich:
@pperiyasamy Good to know that the deployment pod eventually inherits the deployment's namespace, but I still consider it strange that the deployment control loop does not set the namespace initially when it creates the pod in the first place, which, I assume, is what triggers the NRI pod admission controller webhook.

pperiyasamy:
[...]

PR #28, opened 2020-08-05
Handle pods from other kinds properly

pperiyasamy:
This commit makes sure appropriate namespace is retrieved from
pod's ownerrefernce incase if its deployed using other kinds (example:
Deployment/ReplicaSet, DaemonSet, StatefulSet or ReplicationController).
It would enable network-resources-injector to look for network-
attachment-definition object from the correct namespace.

#26 

### DIFF
--- a/deployments/auth.yaml
+++ b/deployments/auth.yaml
@@ -38,6 +38,7 @@ rules:
   - extensions
   - apps
   resources:
+  - replicationcontrollers
   - replicasets
   - daemonsets
   - statefulsets
--- a/deployments/auth.yaml
+++ b/deployments/auth.yaml
@@ -35,7 +35,12 @@ rules:
 - apiGroups:
   - ""
   - k8s.cni.cncf.io
+  - extensions
+  - apps
   resources:
+  - replicasets
+  - daemonsets
+  - statefulsets
   - pods
   - network-attachment-definitions
   verbs:
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -185,6 +185,19 @@ func getNamespaceFromOwnerReference(ownerRef metav1.OwnerReference) (namespace s
 				break
 			}
 		}
+	} else if ownerRef.Kind == "ReplicationController" {
+		var replicationControllers *corev1.ReplicationControllerList
+		replicationControllers, err = clientset.CoreV1().ReplicationControllers("").List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return
+		}
+		for _, replicationController := range replicationControllers.Items {
+			if replicationController.ObjectMeta.Name == ownerRef.Name && replicationController.ObjectMeta.UID == ownerRef.UID {
+				namespace = replicationController.ObjectMeta.Namespace
+				err = nil
+				break
+			}
+		}
 	}
 	if namespace == "" {
 		err = errors.New("pod namespace is not found")
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -24,11 +24,12 @@ import (
 	"strings"
 
 	"github.com/golang/glog"
-	"gopkg.in/intel/multus-cni.v3/types"
 	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/pkg/errors"
+	"gopkg.in/intel/multus-cni.v3/types"
 
 	"k8s.io/api/admission/v1beta1"
+	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


pkg/webhook/webhook.go
```diff
@@ -129,13 +130,40 @@ func deserializePod(ar *v1beta1.AdmissionReview) (corev1.Pod, error) {
 	/* unmarshal Pod from AdmissionReview request */
 	pod := corev1.Pod{}
 	err := json.Unmarshal(ar.Request.Object.Raw, &pod)
-	/* fix for missing "default" namespace */
-	if pod.ObjectMeta.Namespace == "" {
-		pod.ObjectMeta.Namespace = "default"
+	if pod.ObjectMeta.Namespace != "" {
```
JanScheurich, line 133: I think the case that the pod spec does include a Namespace is valid and should be handled normally. We only need to look up the deployment namespace if there is no Namespace at all.

pkg/webhook/webhook.go
```diff
@@ -129,13 +130,68 @@ func deserializePod(ar *v1beta1.AdmissionReview) (corev1.Pod, error) {
...
+		if err != nil {
+			return
+		}
+		for _, statefulSet := range statefulSets.Items {
+			if statefulSet.ObjectMeta.Name == ownerRef.Name && statefulSet.ObjectMeta.UID == ownerRef.UID {
+				namespace = statefulSet.ObjectMeta.Namespace
+				err = nil
+				break
+			}
+		}
```
zshi-redhat, line 187: @pperiyasamy  Could you please add `ReplicationController` as one of the sources to get owner reference? for example:

```
+       } else if ownerRef.Kind == "ReplicationController" {
+               var replicationControllers *corev1.ReplicationControllerList
+               replicationControllers, err = clientset.CoreV1().ReplicationControllers("").List(context.TODO(), metav1.ListOptions{})
+               if err != nil {
+                       return
+               }
+               for _, replicationController := range replicationControllers.Items {
+                       if replicationController.ObjectMeta.Name == ownerRef.Name && replicationController.ObjectMeta.UID == ownerRef.UID {
+                               namespace = replicationController.ObjectMeta.Namespace
+                               err = nil
+                               break
+                       }
+               }
```

This also requires change of ClusterRole in deployments/auth.yaml .

pkg/webhook/webhook.go
```diff
@@ -129,13 +130,81 @@ func deserializePod(ar *v1beta1.AdmissionReview) (corev1.Pod, error) {
...
+			return pod, err
+		}
+		pod.ObjectMeta.Namespace = namespace
 	}
 	return pod, err
 }
 
+func getNamespaceFromOwnerReference(ownerRef metav1.OwnerReference) (namespace string, err error) {
+	namespace = ""
+	if ownerRef.Kind == "ReplicaSet" {
```
ahalim-intel, line 149: I think switch case would be neater construct here instead of bunch of ' if else if'.

pperiyasamy: done.

pkg/webhook/webhook.go
pperiyasamy: yes, done.

zshi-redhat:
@ahalim-intel  @martinkennelly PTAL

deployments/auth.yaml
```diff
@@ -35,7 +35,13 @@ rules:
 - apiGroups:
   - ""
   - k8s.cni.cncf.io
+  - extensions
+  - apps
   resources:
```
przemeklal, line 40: do we need to include deployments in this list as well?

pkg/webhook/webhook.go
```diff
@@ -129,13 +130,85 @@ func deserializePod(ar *v1beta1.AdmissionReview) (corev1.Pod, error) {
...
+			return
+		}
+		for _, replicationController := range replicationControllers.Items {
+			if replicationController.ObjectMeta.Name == ownerRef.Name && replicationController.ObjectMeta.UID == ownerRef.UID {
+				namespace = replicationController.ObjectMeta.Namespace
+				err = nil
+				break
+			}
+		}
+	}
```
przemeklal, line 202: as above, do we need another case for "Deployment" kind?

przemeklal (commented):
In general LGTM, I like that OwnerRef can be used instead of listening for AdmissionReviewRequests for all kinds of applications.

Just 2 questions:
- do we need to have deployments in the list of possible pod owners?
- there's this code duplication across all kinds of possible pod owners, but my guess is that we cannot do too much about it?

pkg/webhook/webhook.go
przemeklal: nvm, just realized that replicaset will be the owner

deployments/auth.yaml
przemeklal: nvm, just realized that replicaset will be the owner

pperiyasamy: The replicaset is the ownerref of pod when its deployed with Deployment kind, The replicaset's ownerref is Deployment. 
Hence this check would not be needed.

pkg/webhook/webhook.go
pperiyasamy: same comment as above.

pperiyasamy:
> In general LGTM, I like that OwnerRef can be used instead of listening for AdmissionReviewRequests for all kinds of applications.
> 
> Just 2 questions:
> 
> * ~do we need to have deployments in the list of possible pod owners?~
> * there's this code duplication across all kinds of possible pod owners, but my guess is that we cannot do too much about it?

I was too thinking about it, but I couldn't find an elegant way to do this, if you've any thoughts, please let me know.

przemeklal:
> > In general LGTM, I like that OwnerRef can be used instead of listening for AdmissionReviewRequests for all kinds of applications.
> > Just 2 questions:
> > 
> > * ~do we need to have deployments in the list of possible pod owners?~
> > * there's this code duplication across all kinds of possible pod owners, but my guess is that we cannot do too much about it?
> 
> I was too thinking about it, but I couldn't find an elegant way to do this, if you've any thoughts, please let me know.

Nothing that would improve it dramatically. Gotta wait for generics to finally arrive in Go I guess :wink:

/lgtm

ahalim-intel:
Thanks @pperiyasamy for this PR and @przemeklal for reviewing

PR #29, opened 2020-08-21
Honor the existing resources quantity

pperiyasamy:
This commit ensures the quantities specified in the resources
are considered along with nri injecting resource based on the
network name annotation.
This way other applications (example: networkservicemesh) can
also request the device(s) from the same resource pool along
with resource requested for multus secondary network interface.

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -39,6 +39,7 @@ func main() {
 	injectHugepageDownApi := flag.Bool("injectHugepageDownApi", false, "Enable hugepage requests and limits into Downward API.")
 	flag.Var(&clientCAPaths, "client-ca", "File containing client CA. This flag is repeatable if more than one client CA needs to be added to server")
 	resourceNameKeys := flag.String("network-resource-name-keys", "k8s.v1.cni.cncf.io/resourceName", "comma separated resource name keys --network-resource-name-keys.")
+	resourceHonorFlag := flag.Bool("honor-resource", false, "Honor the existing resources --honor-resource")
 	flag.Parse()
 
 	if *port < 1024 || *port > 65535 {
@@ -69,6 +70,8 @@ func main() {
 	webhook.SetupInClusterClient()
 
 	webhook.SetInjectHugepageDownApi(*injectHugepageDownApi)
+
+	webhook.SetHonorExistingResources(*resourceHonorFlag)
 
 	err = webhook.SetResourceNameKeys(*resourceNameKeys)
 	if err != nil {
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -39,7 +39,7 @@ func main() {
 	injectHugepageDownApi := flag.Bool("injectHugepageDownApi", false, "Enable hugepage requests and limits into Downward API.")
 	flag.Var(&clientCAPaths, "client-ca", "File containing client CA. This flag is repeatable if more than one client CA needs to be added to server")
 	resourceNameKeys := flag.String("network-resource-name-keys", "k8s.v1.cni.cncf.io/resourceName", "comma separated resource name keys --network-resource-name-keys.")
-	resourceHonorFlag := flag.Bool("honor-resource", false, "Honor the existing resources --honor-resource")
+	resourcesHonorFlag := flag.Bool("honor-resources", false, "Honor the existing requested resources requests & limits --honor-resources")
 	flag.Parse()
 
 	if *port < 1024 || *port > 65535 {
@@ -71,7 +71,7 @@ func main() {
 
 	webhook.SetInjectHugepageDownApi(*injectHugepageDownApi)
 
-	webhook.SetHonorExistingResources(*resourceHonorFlag)
+	webhook.SetHonorExistingResources(*resourcesHonorFlag)
 
 	err = webhook.SetResourceNameKeys(*resourceNameKeys)
 	if err != nil {
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -630,13 +630,18 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
 			glog.Infof("pod doesn't need any custom network resources")
 		} else {
 			var patch []jsonPatchOperation
-
+			var existingrequestsMap map[corev1.ResourceName]resource.Quantity
+			var existingLimitsMap map[corev1.ResourceName]resource.Quantity
 			/* check whether resources paths exists in the first container and add as the first patches if missing */
 			if len(pod.Spec.Containers[0].Resources.Requests) == 0 {
 				patch = patchEmptyResources(patch, 0, "requests")
+			} else {
+				existingrequestsMap = pod.Spec.Containers[0].Resources.Requests.DeepCopy()
 			}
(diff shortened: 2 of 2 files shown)


pperiyasamy:
Is this PR good to go in ? This would enable NRI to interop with other networking solutions (example: networkservicemesh) on the same resource pool.

martinkennelly:
@pperiyasamy I will review this PR now. Thank you for this and your patience. 
@zshi-redhat When you get a chance, could you look at this also?

zshi-redhat:
I'm afraid this will change the default behavior of always injecting the equal number of resources specified in net-attach-defs.
@pperiyasamy if networkservicemesh doesn't need to inject resource name, why NRI is used?

pperiyasamy:
> if networkservicemesh doesn't need to inject resource name, why NRI is used?

@zshi-redhat Thanks for looking at this. Actually both networkservicemesh and [ovs-cni](https://github.com/kubevirt/ovs-cni) (with multus and NRI) work on the same resource pool for our usecase to attach additional  smartnic virtual network interfaces onto the pod container. Yes, NRI injects resources specified in net-attach-defs whereas for networkservicemesh we directly specify the quantities in the pod resource requirements. But NRI replaces it with resources in NAD which makes networkservicemesh not able to find any resources during its network interface plumbing.  Here is an example of such configuration [1](https://github.com/Nordix/networkservicemesh/blob/use-host-ovs/deployments/helm/testpmd-forwarder/ovs-cni-net.yaml#L7), [2](https://github.com/Nordix/networkservicemesh/blob/use-host-ovs/deployments/helm/testpmd-forwarder/templates/testpmd-forwarder-nse.tpl#L16) and [3](https://github.com/Nordix/networkservicemesh/blob/use-host-ovs/deployments/helm/testpmd-forwarder/templates/testpmd-forwarder-nse.tpl#L52). This PR would eventually fixes the problem.

zshi-redhat:
> > if networkservicemesh doesn't need to inject resource name, why NRI is used?
> 
> @zshi-redhat Thanks for looking at this. Actually both networkservicemesh and [ovs-cni](https://github.com/kubevirt/ovs-cni) (with multus and NRI) work on the same resource pool for our usecase to attach additional smartnic virtual network interfaces onto the pod container. Yes, NRI injects resources specified in net-attach-defs whereas for networkservicemesh we directly specify the quantities in the pod resource requirements. But NRI replaces it with resources in NAD which makes networkservicemesh not able to find any resources during its network interface plumbing. Here is an example of such configuration [1](https://github.com/Nordix/networkservicemesh/blob/use-host-ovs/deployments/helm/testpmd-forwarder/ovs-cni-net.yaml#L7), [2](https://github.com/Nordix/networkservicemesh/blob/use-host-ovs/deployments/helm/testpmd-forwarder/templates/testpmd-forwarder-nse.tpl#L16) and [3](https://github.com/Nordix/networkservicemesh/blob/use-host-ovs/deployments/helm/testpmd-forwarder/templates/testpmd-forwarder-nse.tpl#L52). This PR would eventually fixes the problem.

I don't have much experience about networkservicemesh, maybe the right approach is to have network-resources-injector  recognize the networkserviemesh `annotation` (or whatever that express the resource request by servicemsh) when injecting resourceNames?

pperiyasamy:
> network-resources-injector recognize the networkserviemesh `annotation` (or whatever that express the resource request by servicemsh) when injecting resourceNames

afaik, the current nsm doesn't provide a way to request a resource via `annotation`, currently we either have this NRI fix or use different resource pools (using VF index) to solve the issue.

martinkennelly:
@pperiyasamy Is it possible to rework this to inject NAD requests/limits to ensure we get equal number of resources injected as it is right now, but also honor existing requests/limits on a best effort basis? If there is a resource injection conflict NAD annotation injected resources wins. Don't go ahead with this work until @zshi-redhat has had input on this.

zshi-redhat:
> @pperiyasamy Is it possible to rework this to inject NAD requests/limits to ensure we get equal number of resources injected as it is right now, but also honor existing requests/limits on a best effort basis? If there is a resource injection conflict NAD annotation injected resources wins. Don't go ahead with this work until @zshi-redhat has had input on this.

Yes, I think what @martinkennelly suggested might be a intermediate solution that NRI overwrites resource request/limit as a default behavior and respect existing resource request/limit by enabling a flag. The reason why we need a flag is that I don't see NRI can predict whether user wants it to overwrite or append, wdyt?

pperiyasamy:
@zshi-redhat @martinkennelly Thanks for your feedback, yes, Let's go ahead with a flag called `honorExistingResources `which is set to false by default. I'll make the required changes and push an another commit.

pkg/webhook/webhook.go
```diff
@@ -594,31 +668,10 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
...
-					Operation: "add",
-					Path:      "/spec/containers/0/resources/requests/" + toSafeJsonPatchKey(resource.String()), // NOTE: in future we may want to patch specific container (not always the first one)
-					Value:     quantity,
-				})
-				patch = append(patch, jsonPatchOperation{
-					Operation: "add",
-					Path:      "/spec/containers/0/resources/limits/" + toSafeJsonPatchKey(resource.String()),
-					Value:     quantity,
-				})
+			if honorExistingResources {
```
martinkennelly, line 671: can you add a ```glog.Infof("honor-resource=%v", honorExistingResources)``` before the if statement?

cmd/webhook/main.go
```diff
@@ -39,6 +39,7 @@ func main() {
 	injectHugepageDownApi := flag.Bool("injectHugepageDownApi", false, "Enable hugepage requests and limits into Downward API.")
 	flag.Var(&clientCAPaths, "client-ca", "File containing client CA. This flag is repeatable if more than one client CA needs to be added to server")
 	resourceNameKeys := flag.String("network-resource-name-keys", "k8s.v1.cni.cncf.io/resourceName", "comma separated resource name keys --network-resource-name-keys.")
+	resourceHonorFlag := flag.Bool("honor-resource", false, "Honor the existing resources --honor-resource")
```
martinkennelly, line 42: I am inclined to put the plural here. ```--honor-resources``` and also expand the description to say "Honor the existing requested resources requests & limits --honor-resources"

martinkennelly (commented):
Tested it and works as expected. Slightly concerned about code duplication but overall looks good. Waiting on @zshi-redhat input. Thank you for this!

zshi-redhat:
looks good to me if Martin's comments are addressed.

pkg/webhook/webhook.go
pperiyasamy: done

cmd/webhook/main.go
pperiyasamy: sure, done

pperiyasamy:
> Tested it and works as expected. Slightly concerned about code duplication but overall looks good. Waiting on @zshi-redhat input. Thank you for this!

@martinkennelly thanks, Addressed the code duplication as much as possible, hope it's ok now.

martinkennelly (approved):
Thanks for your patience with this patch Peri!

PR #31, opened 2020-08-24
add default handler for getting ownerReference

zshi-redhat:
Without default handler, network-resources-injector rejects the pod
creation from any unsupported Kind.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -199,6 +199,10 @@ func getNamespaceFromOwnerReference(ownerRef metav1.OwnerReference) (namespace s
 				break
 			}
 		}
+	default:
+		glog.Infof("owner reference kind is not supported: %v, using default namespace", ownerRef.Kind)
+		namespace = "default"
+		return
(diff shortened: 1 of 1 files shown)


pkg/webhook/webhook.go
```diff
@@ -199,6 +199,9 @@ func getNamespaceFromOwnerReference(ownerRef metav1.OwnerReference) (namespace s
 				break
 			}
 		}
+	default:
+		glog.Infof("owner reference kind is not supported: %v", ownerRef.Kind)
+		return
```
pperiyasamy, line 204: shouldn't namespace set to default before returning it ?

zshi-redhat: I was thinking there is not much different return empty or default namespace, but I think at least it can cover the case where pod is created in default namespace. I will make the change.

zshi-redhat: Done

martinkennelly:
/lgtm
Thanks for the patch!

PR #33, opened 2020-09-25
Correct static scan issues

martinkennelly:


martinkennelly:

pkg/webhook/webhook.go
```diff
@@ -500,7 +500,7 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
 			}
 
 			patch = createVolPatch(patch)
-			glog.Infof("patch after all mutations", patch)
+			glog.Infof("patch after all mutations: %v", patch)
```
zshi-redhat, line 503: good catch

scripts/webhook-deployment.sh
```diff
@@ -1,10 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 # Always exit on errors.
 set -e
 
 # Set our known directories and parameters.
-BASE_DIR=$(cd $(dirname $0)/..; pwd)
+BASE_DIR="$(cd "$(dirname "$0")"/..; pwd)"
```
zshi-redhat, line 21: we have double quotes inside double quotes, do we need escape? or should it just work?

Dockerfile
```diff
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 FROM golang:1.13-alpine as builder
-ADD . /usr/src/network-resources-injector
-RUN apk add --update --virtual build-dependencies build-base linux-headers && \
-    cd /usr/src/network-resources-injector && \
+COPY . /usr/src/network-resources-injector
```
zshi-redhat, line 16: @martinkennelly I have seen people using ADD and COPY to achieve the same copy task., but don't know which one is better over the other in a certain situation.  which tool are you using to check the dockerfile semantics?

scripts/webhook-deployment.sh
martinkennelly: It works fine this way with bash. I have tested it. You can have nested double quotes when using $().

Dockerfile
martinkennelly: I used hadolint. It will also inform you why something is incorrect. You can use snyk also.
https://github.com/hadolint/hadolint

zshi-redhat:
/lgtm
@ahalim-intel @martinkennelly I don't have merge permission to this repo, would one of you merge it?

PR #35, opened 2020-09-25
Harden TLS endpoint security

martinkennelly:
Please see commit messages for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,3 +113,11 @@ To create the vendor folder invoke the following which will create a vendor fold
 ```bash
 make vendor
 ```
+
+## Security
+### Disable adding client CAs to server TLS endpoint
+If you wish to not add any client CAs to the servers TLS endpoint, add ```--insecure``` flag to webhook binary arguments.
+
+### Client CAs
+By default, we consume the client CA from the Kubernetes service account secrets directory ```/var/run/secrets/kubernetes.io/serviceaccount/```.
+If you wish to consume a client CA from a different location, please specify flag ```--client-ca``` with a valid path. If you wish to add more than one client CA, repeat this flag multiple times. If ```--client-ca``` is defined, the default client CA from the service account secrets directory will not be consumed.
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/golang/glog"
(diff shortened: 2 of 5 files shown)


martinkennelly:
@zshi-redhat Could you review this please?

martinkennelly:
@zshi-redhat Hey, thank you for your previous reviews. Could you look at this when you have bandwidth? Just a reminder incase it slipped off your workload. This is the most important commit series, so I understand the need to test this sufficiently on your side.

zshi-redhat:
> @zshi-redhat Hey, thank you for your previous reviews. Could you look at this when you have bandwidth? Just a reminder incase it slipped off your workload. This is the most important commit series, so I understand the need to test this sufficiently on your side.

@martinkennelly thanks for reminding!
I tried last week but failed to build a image for testing. It might be due to some mis-configuration in my environment. 
The problem I had was I cannot build the image using alpine base image. I then tried to using centos:7 base, but failed to resolve the go pkg dependencies.
I will try some other approaches this week.

martinkennelly:
@zshi-redhat I just tested building this PR on top of existing NRI master on a fresh machine here in Intel and it compiles as expected.

zshi-redhat:
> @zshi-redhat I just tested building this PR on top of existing NRI master on a fresh machine here in Intel and it compiles as expected.

OK, my issue was I always got `DNS lookup error` when building with alpine based image. and with default golang-1.13 in centos:7 base image, it failed to resolve pkg dependency. I switched to use centos:7 + golang-1.15, the build succeeded.

I was able to test this PR successfully on a openshift environment with `--insecure` flag specified.
I'm not very familiar with these newly added tls options, but I think when `--insecure` is used, the major change for added client-ca is disabled, I may come back with question on other tls options if later a problem is hit, so far it looks good to me.

one nit: Since `--insecure` is to a control switch for `--client-ca` and it doesn't exclude us from using default ca or self-provided ca (by attaching a secret to network-resources-injector daemonset), would it make sense to set `--insecure` automatically inside webhook when number of  specified `--client-ca` is non-zero?

martinkennelly:
Existing behavior is only maintained by specifying insecure flag with this current patch. I think its easier to understand this current patches flow with a flow chart diagram [here](https://imgur.com/XYwOr0M).

Your question has highlighted that this is now confusing and needs more documentation or simplification.

What could be done is to remove insecure flag entirely and remove adding the default client CA (from /var/run/secrets...)  when there is no ```--client-ca's``` flags defined like [this](https://imgur.com/a/IyuguQo). Let the user add the default client CA via ```--client-ca``` flag if need be. What you think?

zshi-redhat:
> @zshi-redhat Existing behavior is only maintained by specifying insecure flag with this current patch. I think its easier to understand this current patches flow with a flow chart diagram [here](https://imgur.com/XYwOr0M).
> 
> Your question has highlighted that this is now confusing and needs more documentation or simplification.
> 
> What could be done is to remove insecure flag entirely and remove adding the default client CA (from /var/run/secrets...) when there is no `--client-ca's` flags defined like [this](https://imgur.com/a/IyuguQo). Let the user add the default client CA via `--client-ca` flag if need be. What you think?

ok, I missed the default CA. I was thinking to set --insecure automatically when number of --client-ca is non-zero, but it looks like we could always use the default CA.
given that, I don't think we need to change current cmd options.

/lgtm

PR #37, opened 2020-09-25
Add failure if parsing networks failed

martinkennelly:
* Return bad request if we cannot parse network
  annotation

* Previous to this patch, it was ignored and pod would
  be scheduled

Signed-off-by: Kennelly, Martin <martin.kennelly@intel.com>


martinkennelly:

zshi-redhat:
@ahalim-intel @martinkennelly I don't have merge permission to this repo, would one of you merge it?

PR #41, opened 2020-11-20
Remove directories .gopath & .idea

### LINKED ISSUE #40: Remove .idea & .gopath
### What happened?
Both directories .idea and .gopath should not be in the code base. .idea is specific to a jetbrains product and .gopath is no longer needed due to go mod support

### What did you expect to happen?
NA
### What are the minimal steps needed to reproduce the bug?
NA
### Anything else we need to know?
Will be corrected in PR.

martinkennelly:
Both of these directories are not required to be in the code base. .gopath directory is no longer needed with go mod support.
Fixes #40


killianmuldoon (approved):
/LGTM

PR #42, opened 2020-12-01
Expose hugepage requests to container via Downward API

Billy99:
In Kubernetes 1.20, an alpha feature was added to expose the requested
hugepages to the container via the Downward API. This is feature is
disabled by default. If enabled when Kubernetes is deployed via
FEATURE_GATES="DownwardAPIHugePages=true", then Network Resource
Injector can be used to mutate the pod spec to publish the hugepage
data to the container. Since the Kubernetes feature is disabled by
default, added a configuration parameter to turn on the Network
Resource Injector code.

NOTE: The code loops through all containers in a pod and if multiple
containers are requesting hugepages, then the Downward API fields
are requested for each containers. There is an issue where the same
value is placed in both containers, even if different values are requested.
Think it is a K8s issue and still investigating.

Signed-off-by: Billy McFall <22157057+Billy99@users.noreply.github.com>

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -1,0 +1,25 @@
+// Copyright (c) 2020 Red Hat, Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+const (
+	DownwardAPIMountPath   = "/etc/podnetinfo"
+	AnnotationsPath        = "annotations"
+	LabelsPath             = "labels"
+	Hugepages1GRequestPath = "hugepages_1G_request"
+	Hugepages2MRequestPath = "hugepages_2M_request"
+	Hugepages1GLimitPath   = "hugepages_1G_limit"
+	Hugepages2MLimitPath   = "hugepages_2M_limit"
+)
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -74,7 +74,7 @@ func readAdmissionReview(req *http.Request, w http.ResponseWriter) (*v1beta1.Adm
 	var body []byte
 
 	if req.Body != nil {
-		req.Body = http.MaxBytesReader(w, req.Body, 1 << 20)
+		req.Body = http.MaxBytesReader(w, req.Body, 1<<20)
 		if data, err := ioutil.ReadAll(req.Body); err == nil {
 			body = data
 		}
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -26,8 +26,9 @@ import (
 	"github.com/golang/glog"
 	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/pkg/errors"
-	"gopkg.in/intel/multus-cni.v3/types"
+	multus "gopkg.in/intel/multus-cni.v3/types"
 
+	"github.com/intel/network-resources-injector/pkg/types"
 	"k8s.io/api/admission/v1beta1"
 	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -45,13 +46,20 @@ type jsonPatchOperation struct {
 	Value     interface{} `json:"value,omitempty"`
 }
 
+type hugepageResourceData struct {
+	ResourceName  string
+	ContainerName string
+	Path          string
+}
+
 const (
 	networksAnnotationKey = "k8s.v1.cni.cncf.io/networks"
 )
 
 var (
-	clientset        kubernetes.Interface
-	resourceNameKeys []string
+	clientset             kubernetes.Interface
+	injectHugepageDownApi bool
+	resourceNameKeys      []string
 )
 
 func prepareAdmissionReviewResponse(allowed bool, message string, ar *v1beta1.AdmissionReview) error {
@@ -220,8 +228,8 @@ func toSafeJsonPatchKey(in string) string {
 	return out
 }
 
-func parsePodNetworkSelections(podNetworks, defaultNamespace string) ([]*types.NetworkSelectionElement, error) {
-	var networkSelections []*types.NetworkSelectionElement
+func parsePodNetworkSelections(podNetworks, defaultNamespace string) ([]*multus.NetworkSelectionElement, error) {
+	var networkSelections []*multus.NetworkSelectionElement
 
 	if len(podNetworks) == 0 {
 		err := errors.New("empty string passed as network selection elements list")
@@ -257,9 +265,9 @@ func parsePodNetworkSelections(podNetworks, defaultNamespace string) ([]*types.N
 	return networkSelections, nil
 }
 
-func parsePodNetworkSelectionElement(selection, defaultNamespace string) (*types.NetworkSelectionElement, error) {
+func parsePodNetworkSelectionElement(selection, defaultNamespace string) (*multus.NetworkSelectionElement, error) {
 	var namespace, name, netInterface string
-	var networkSelectionElement *types.NetworkSelectionElement
+	var networkSelectionElement *multus.NetworkSelectionElement
 
 	units := strings.Split(selection, "/")
 	switch len(units) {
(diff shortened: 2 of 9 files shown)


Billy99:
@zshi-redhat PTAL
First pass. Let me know if you ahve any issues.

pkg/webhook/webhook.go
```diff
@@ -45,13 +46,20 @@ type jsonPatchOperation struct {
 	Value     interface{} `json:"value,omitempty"`
 }
 
+type hugepageResourceData struct {
+	ContainerName string
+    Path          string
```
zshi-redhat, line 51: indent doesn't match with above line.

zshi-redhat (commented):
Billy, just one minor syntax comment.

btw, do you have an example or reference on what the pod spec looks like after injecting hugepage downward API?

Billy99:
> btw, do you have an example or reference on what the pod spec looks like after injecting hugepage downward API?

Not easiest to read, but from the log:
{"log":"I1201 00:14:07.848642       1 webhook.go:571] patch after all mutations: [{add /spec/containers/0/resources/requests/intel.com~1intel_sriov_dpdk_a {{1 0} {\u003cnil\u003e}  DecimalSI}} {add /spec/containers/0/resources/limits/intel.com~1intel_sriov_dpdk_a {{1 0} {\u003cnil\u003e}  DecimalSI}} {add /spec/containers/0/resources/requests/intel.com~1intel_sriov_dpdk_b {{1 0} {\u003cnil\u003e}  DecimalSI}} {add /spec/containers/0/resources/limits/intel.com~1intel_sriov_dpdk_b {{1 0} {\u003cnil\u003e}  DecimalSI}} {add /spec/containers/0/volumeMounts/- {podnetinfo false /etc/podnetinfo  \u003cnil\u003e }} {add /spec/volumes/- {podnetinfo {nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil \u0026DownwardAPIVolumeSource{Items:[]DownwardAPIVolumeFile{DownwardAPIVolumeFile{Path:labels,FieldRef:\u0026ObjectFieldSelector{APIVersion:,FieldPath:metadata.labels,},ResourceFieldRef:nil,Mode:nil,},DownwardAPIVolumeFile{Path:annotations,FieldRef:\u0026ObjectFieldSelector{APIVersion:,FieldPath:metadata.annotations,},ResourceFieldRef:nil,Mode:nil,},DownwardAPIVolumeFile{Path:hugepages_1Gi_request,FieldRef:nil,ResourceFieldRef:\u0026ResourceFieldSelector{ContainerName:sriov-example,Resource:requests.hugepages-1Gi,Divisor:{{1073741824 0} {\u003cnil\u003e}  BinarySI},},Mode:nil,},DownwardAPIVolumeFile{Path:hugepages_1Gi_limit,FieldRef:nil,ResourceFieldRef:\u0026ResourceFieldSelector{ContainerName:sriov-example,Resource:limits.hugepages-1Gi,Divisor:{{1073741824 0} {\u003cnil\u003e}  BinarySI},},Mode:nil,},},DefaultMode:nil,} nil nil nil nil nil nil nil nil nil nil nil nil}}}]\n","stream":"stderr","time":"2020-12-01T00:14:07.848890482Z"}

pkg/types/types.go
```diff
@@ -0,0 +1,25 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+const (
+	Hugepage1GiRequestPath = "hugepages_1Gi_request"
```
martinkennelly, line 18: I believe these paths are used to create the downward API file. Are these paths standardized in the k8 community somewhere? Should we be pulling in these paths from somewhere instead of defining them ourselves?

pkg/webhook/webhook.go
```diff
@@ -504,7 +526,48 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
...
+			hugepageResourceList := make(map[string]hugepageResourceData)
+			glog.Infof("injectHugepageDownApi=%v", injectHugepageDownApi)
+			if injectHugepageDownApi {
+				for _, container := range pod.Spec.Containers {
+					if len(container.Resources.Requests) != 0 {
+						if quantity, exists := container.Resources.Requests["hugepages-1Gi"]; exists && quantity.IsZero() == false {
+							hugepageResourceList["requests.hugepages-1Gi"] = hugepageResourceData{
+								ContainerName: container.Name,
+								Path:          types.Hugepage1GiRequestPath,
+								DivisorVal:    1*1024*1024*1024,
```
martinkennelly: How is the quantity of the request exposed to pod? Number of 1Gi huge pages for example? I don't see it here.

pkg/webhook/webhook.go
Billy99: Fixed

Billy99:
@martinkennelly PTAL

Removed the WIP. I need to test 2M hugepages, but other than verifying that, I think it is ready.

martinkennelly:
> @martinkennelly PTAL
> 
> Removed the WIP. I need to test 2M hugepages, but other than verifying that, I think it is ready.

I am testing it myself also right now.

pkg/webhook/webhook.go
Billy99: It will show up as it is reserved. For 1G hugepages for example. where 2G has been requested:
```
sh-4.2# cat /etc/podnetinfo/req_hugepages_1Gi 
2
sh-4.2# cat /etc/podnetinfo/lim_hugepages_1Gi 
2
```
Need to make sure 2M is correct.

Billy99: 2M works as coded, for 2G at 2M hugepage size,
```
sh-4.2# cat /etc/podnetinfo/hugepages_2Mi_request 
2000
sh-4.2# cat /etc/podnetinfo/hugepages_2Mi_limit
2000
```
However, now that you have me thinking about it, the container doesn't care if it is 1G or 2M hugepage size, DPDK for example just uses `-m 1024`. So I am going to slightly rework so that the value is constant and can just be applied. Also, see pending reply to your other comment about standard path.

pkg/types/types.go
Billy99: @martinkennelly 
I can't find a standard path name in the Kubernetes code, which is why I added a `pkg/types/` so the path name could be reused in app-netutil. I even looked for others, like `cpu_limit` and `mem_request`. We don't even use the standard path of `/etc/podinfo/`. We use `/etc/podnetinfo/` to avoid any conflicts.

I used https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/ as my naming convention. Digging though the hugepage patch more (https://github.com/kubernetes/kubernetes/pull/86102/files), in  pkg/apis/core/validation/validation_test.go (was hidden because diff was too large), I see the test code uses `hugepages_request` and `hugepages_limit`. Kubernetes doesn't support mixing 1G and 2M hugepages, so should I just move to that convention? Or be forward looking and leave as is in case it is ever supported?

Billy99: Whatever naming convention I use, I could move the constants somewhere else, like `k8snetworkplumbingwg/network-attachment-definition-client`? May slow this patch down slightly.

Billy99: Latest commit change to use `hugepages_request` and `hugepages_limit`.

martinkennelly: I believe you can have both 1G and 2M hugepages for a pod ( https://kubernetes.io/docs/tasks/manage-hugepages/scheduling-hugepages/#api )

Billy99: @martinkennelly OK, I'll change to allow both. Any other concerns?

martinkennelly: _IF_ I am right about being able to use both 2m/1g hugepages at the same time, then there is a problem here. You've no way for an app to distinguish between both first off, but even then, only one will be selected I guess and populate hugepages_request/hugepages_limit. I must test this out. I haven't found docs to say I can use both yet.

Billy99: Year and half ago, my server was setup with both 1G and 2M hugpages and K8s wouldn't deploy. I found somewhere that indicated it wasn't supported. But a lot has change since then.

martinkennelly: I believe you can with this feature gate: --feature-gates=HugePageStorageMediumSize=true

Billy99: OK, I'll add _1G and _2M back to the "Path"  name.

zshi-redhat:
> > @martinkennelly PTAL
> > Removed the WIP. I need to test 2M hugepages, but other than verifying that, I think it is ready.
> 
> I am testing it myself also right now.

@martinkennelly 
Talking to Billy, I think we are good with having this hugepage downward API support in general so I'd like to merge it first in favor of a downstream backport to meet the timeline of feature inclusion.
Sorry for having to rush merging on it.
We'd love to fix any issue you may find in separate PRs.

zshi-redhat (approved):

Issue #45, opened 2020-12-03
Downward API Injection: Hugepage injection doesn't work with mult container


Billy99:
On top of Issue #44 (Downward API Injection: Multiple containers in pod don't work), when manually adding hugepage fields to the Downward API for multiple containers, the same value is reported in each container, even if the requests are different. This is most likely a Kubernetes issue (will update with issue link when issue is found), but tracking here for documentation of issue.

Issue #44, opened 2020-12-03
Downward API Injection: Multiple containers in pod don't work


Billy99:
If there multiple containers in a pod that need Downward API fields injected, it doesn't work. There is logic in code that only looks at `pod.Spec.Containers[0]`.

pperiyasamy:
@Billy99 Can I work on a PR for injecting podnetinfo into multiple containers ? Please let me know.

Billy99:
@pperiyasamy  Yes, I have not started and won't have time in the near term. While you are looking at logic to walk through the list of containers, may be easy to add logic to address #43 as well (but your call).

pperiyasamy:
@Billy99 since PR #51 is merged now, can we close this issue ?

Billy99:
@pperiyasamy Yes, tested and looks good.

PR #46, opened 2020-12-08
Add container name to Hugepage Downward API file names

Billy99:
To support multiple containers in a pod using hugepages, append the container
name to the Hugepage Downward API file names.

Resolves: #45

Signed-off-by: Billy McFall <22157057+Billy99@users.noreply.github.com>

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -429,6 +429,58 @@ func createVolPatch(patch []jsonPatchOperation, hugepageResourceList []hugepageR
 	return patch
 }
 
+func addEnvVar(patch []jsonPatchOperation, containerIndex int, firstElement bool,
+	envName string, envVal string) []jsonPatchOperation {
+
+	env := corev1.EnvVar{
+		Name:  envName,
+		Value: envVal,
+	}
+
+	if firstElement {
+		patch = append(patch, jsonPatchOperation{
+			Operation: "add",
+			Path:      "/spec/containers/" + fmt.Sprintf("%d", containerIndex) + "/env",
+			Value:     []corev1.EnvVar{env},
+		})
+	} else {
+		patch = append(patch, jsonPatchOperation{
+			Operation: "add",
+			Path:      "/spec/containers/" + fmt.Sprintf("%d", containerIndex) + "/env/-",
+			Value:     env,
+		})
+	}
+
+	return patch
+}
+
+func createEnvPatch(patch []jsonPatchOperation, container *corev1.Container,
+	containerIndex int, envName string, envVal string) []jsonPatchOperation {
+
+	// Determine if requested ENV already exists
+	found := false
+	firstElement := false
+	if len(container.Env) != 0 {
+		for _, env := range container.Env {
+			if env.Name == envName {
+				found = true
+				if env.Value != envVal {
+					glog.Warningf("Error, adding env '%s', name existed but value different: '%s' != '%s'",
+						envName, env.Value, envVal)
+				}
+				break
+			}
+		}
+	} else {
+		firstElement = true
+	}
+
+	if !found {
+		patch = addEnvVar(patch, containerIndex, firstElement, envName, envVal)
+	}
+	return patch
+}
+
 // MutateHandler handles AdmissionReview requests and sends responses back to the K8s API server
 func MutateHandler(w http.ResponseWriter, req *http.Request) {
 	glog.Infof("Received mutation request")
@@ -531,23 +583,26 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
 			var hugepageResourceList []hugepageResourceData
 			glog.Infof("injectHugepageDownApi=%v", injectHugepageDownApi)
 			if injectHugepageDownApi {
-				for _, container := range pod.Spec.Containers {
+				for containerIndex, container := range pod.Spec.Containers {
+					found := false
 					if len(container.Resources.Requests) != 0 {
 						if quantity, exists := container.Resources.Requests["hugepages-1Gi"]; exists && quantity.IsZero() == false {
 							hugepageResource := hugepageResourceData{
 								ResourceName:  "requests.hugepages-1Gi",
 								ContainerName: container.Name,
-								Path:          types.Hugepages1GRequestPath,
+								Path:          types.Hugepages1GRequestPath + "_" + container.Name,
 							}
 							hugepageResourceList = append(hugepageResourceList, hugepageResource)
+							found = true
 						}
 						if quantity, exists := container.Resources.Requests["hugepages-2Mi"]; exists && quantity.IsZero() == false {
 							hugepageResource := hugepageResourceData{
 								ResourceName:  "requests.hugepages-2Mi",
 								ContainerName: container.Name,
-								Path:          types.Hugepages2MRequestPath,
+								Path:          types.Hugepages2MRequestPath + "_" + container.Name,
 							}
 							hugepageResourceList = append(hugepageResourceList, hugepageResource)
+							found = true
 						}
 					}
 					if len(container.Resources.Limits) != 0 {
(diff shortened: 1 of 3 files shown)


Billy99:
@zshi-redhat @martinkennelly PTAL

zshi-redhat:
@Billy99 would the application knows the container name where it is running?

Billy99:
> @Billy99 would the application knows the container name where it is running?

@zshi-redhat I haven't found an automated way yet. This PR and the app-netutil would have gone in last week if I had. The only solution I have found is to add an environment variable with the container name. See: https://github.com/openshift/app-netutil/pull/28/commits/a0259978a1a245797b141424d3086444782ee1d0#diff-9781b7146e794401aae854f461a50bdf9623bbac06dc9f97e05d582400e1e02bR40-R42

zshi-redhat:
> > @Billy99 would the application knows the container name where it is running?
> 
> @zshi-redhat I haven't found an automated way yet. This PR and the app-netutil would have gone in last week if I had. The only solution I have found is to add an environment variable with the container name. See: [openshift/app-netutil@a025997#diff-9781b7146e794401aae854f461a50bdf9623bbac06dc9f97e05d582400e1e02bR40-R42](https://github.com/openshift/app-netutil/commit/a0259978a1a245797b141424d3086444782ee1d0#diff-9781b7146e794401aae854f461a50bdf9623bbac06dc9f97e05d582400e1e02bR40-R42)

Do you think we shall pass container name to app so that it knows which hugepage is mapped to which container?

Billy99:
> > > @Billy99 would the application knows the container name where it is running?
> > 
> > 
> > @zshi-redhat I haven't found an automated way yet. This PR and the app-netutil would have gone in last week if I had. The only solution I have found is to add an environment variable with the container name. See: [openshift/app-netutil@a025997#diff-9781b7146e794401aae854f461a50bdf9623bbac06dc9f97e05d582400e1e02bR40-R42](https://github.com/openshift/app-netutil/commit/a0259978a1a245797b141424d3086444782ee1d0#diff-9781b7146e794401aae854f461a50bdf9623bbac06dc9f97e05d582400e1e02bR40-R42)
> 
> Do you think we shall pass container name to app so that it knows which hugepage is mapped to which container?

@zshi-redhat Done. PTAL

zshi-redhat:
> > > > @Billy99 would the application knows the container name where it is running?
> > > 
> > > 
> > > @zshi-redhat I haven't found an automated way yet. This PR and the app-netutil would have gone in last week if I had. The only solution I have found is to add an environment variable with the container name. See: [openshift/app-netutil@a025997#diff-9781b7146e794401aae854f461a50bdf9623bbac06dc9f97e05d582400e1e02bR40-R42](https://github.com/openshift/app-netutil/commit/a0259978a1a245797b141424d3086444782ee1d0#diff-9781b7146e794401aae854f461a50bdf9623bbac06dc9f97e05d582400e1e02bR40-R42)
> > 
> > 
> > Do you think we shall pass container name to app so that it knows which hugepage is mapped to which container?
> 
> @zshi-redhat Done. PTAL

LGTM

zshi-redhat (approved):

pperiyasamy:
@Billy99 does this fix the following issue? i could see a problem even with pod having single container requesting 1Gi hugepages.

```
# kubectl create -f deployments/pod-tc1.yaml
The Pod "testpod1" is invalid:
* spec.volumes[2].downwardAPI.resourceFieldRef.resource: Unsupported value: "requests.hugepages-1Gi": supported values: "limits.cpu", "limits.ephemeral-storage", "limits.memory", "requests.cpu", "requests.ephemeral-storage", "requests.memory"
* spec.volumes[2].downwardAPI.resourceFieldRef.resource: Unsupported value: "limits.hugepages-1Gi": supported values: "limits.cpu", "limits.ephemeral-storage", "limits.memory", "requests.cpu", "requests.ephemeral-storage", "requests.memory"
* spec.containers[0].volumeMounts[2].name: Not found: "podnetinfo"
```

Billy99:
> @Billy99 does this fix the following issue? i could see a problem even with pod having single container requesting 1Gi hugepages.
> 
> ```
> # kubectl create -f deployments/pod-tc1.yaml
> The Pod "testpod1" is invalid:
> * spec.volumes[2].downwardAPI.resourceFieldRef.resource: Unsupported value: "requests.hugepages-1Gi": supported values: "limits.cpu", "limits.ephemeral-storage", "limits.memory", "requests.cpu", "requests.ephemeral-storage", "requests.memory"
> * spec.volumes[2].downwardAPI.resourceFieldRef.resource: Unsupported value: "limits.hugepages-1Gi": supported values: "limits.cpu", "limits.ephemeral-storage", "limits.memory", "requests.cpu", "requests.ephemeral-storage", "requests.memory"
> * * spec.volumes[2].downwardAPI.resourceFieldRef.resource: Unsupported value: "requests.hugepages-1Gi": supported values: "limits.cpu", "limits.ephemeral-storage", "limits.memory", "requests.cpu", "requests.ephemeral-storage", "requests.memory"
> ```
No, my initial guess is that "DownwardAPIHugePages" an alpha feature and alpha features are disabled on K8s by default. To use these Downward API fields, you need K8s 1.20 and "DownwardAPIHugePages"  Feature Gate enabled. I could not find any way from within the code to poll K8s to determine if a Feature Gate is supported and enabled. So the design is to enable the feature in NRI via the `--injectHugepageDownApi` parameter and SR-IOV operator will only enable if in a K8s 1.20 and "DownwardAPIHugePages"  Feature Gate enabled deployment.

pperiyasamy:
> > @Billy99 does this fix the following issue? i could see a problem even with pod having single container requesting 1Gi hugepages.
> > ```
> > # kubectl create -f deployments/pod-tc1.yaml
> > The Pod "testpod1" is invalid:
> > * spec.volumes[2].downwardAPI.resourceFieldRef.resource: Unsupported value: "requests.hugepages-1Gi": supported values: "limits.cpu", "limits.ephemeral-storage", "limits.memory", "requests.cpu", "requests.ephemeral-storage", "requests.memory"
> > * spec.volumes[2].downwardAPI.resourceFieldRef.resource: Unsupported value: "limits.hugepages-1Gi": supported values: "limits.cpu", "limits.ephemeral-storage", "limits.memory", "requests.cpu", "requests.ephemeral-storage", "requests.memory"
> > * * spec.volumes[2].downwardAPI.resourceFieldRef.resource: Unsupported value: "requests.hugepages-1Gi": supported values: "limits.cpu", "limits.ephemeral-storage", "limits.memory", "requests.cpu", "requests.ephemeral-storage", "requests.memory"
> > ```
> 
> @pperiyasamy No, my initial guess is that "DownwardAPIHugePages" an alpha feature and alpha features are disabled on K8s by default. To use these Downward API fields, you need K8s 1.20 and "DownwardAPIHugePages" Feature Gate enabled. I could not find any way from within the code to poll K8s to determine if a Feature Gate is supported and enabled. So the design is to enable the feature in NRI via the `--injectHugepageDownApi` parameter and SR-IOV operator will only enable if in a K8s 1.20 and "DownwardAPIHugePages" Feature Gate enabled deployment.

@Billy99 ah! ok, I understand now, I'm still with k8s 1.19.3, will try with K8s 1.20 DownwardAPIHugePages feature gate enabled. thanks.

pkg/webhook/webhook.go
```diff
@@ -429,6 +429,58 @@ func createVolPatch(patch []jsonPatchOperation, hugepageResourceList []hugepageR
...
+
+	// Determine if requested ENV already exists
+	found := false
+	firstElement := false
+	if len(container.Env) != 0 {
+		for _, env := range container.Env {
+			if env.Name == envName {
+				found = true
+				if env.Value != envVal {
+					glog.Infof("Error, adding env %s, name existed but value different: %s != %s",
```
martinkennelly, line 468: I see there are three format specifiers here (%s x 3) and two values, is this ok? Also, I really like enclosing the specifiers in single quotes as well but not necessary here.
Nit: Also, do you think this should be a warning instead of info?

martinkennelly (commented):
Looks good Billy. I am now testing it.

pkg/webhook/webhook.go
```diff
@@ -531,43 +583,56 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
 			var hugepageResourceList []hugepageResourceData
 			glog.Infof("injectHugepageDownApi=%v", injectHugepageDownApi)
 			if injectHugepageDownApi {
-				for _, container := range pod.Spec.Containers {
+				for containerIndex, container := range pod.Spec.Containers {
```
pperiyasamy, line 586: Look the resource patch creation logic still considers pod with single container. shouldn't we fix that too for multiple containers ?

Billy99: I created Issue #44 to track it. Looks like the problem has been in  the code for a while. Not sure I have time at the moment to fix, but it definitely needs to be fixed. I created a podspec manually to make sure the hugepages worked in the multi-container environment.

pkg/webhook/webhook.go
Billy99: @martinkennelly  Fixed. Thanks.

martinkennelly: I was thinking more about using glog.Warningf. glog.info could get filtered out.

martinkennelly:
Can you add to the readme about this multi container support and expected path & labels?
Something like:


Hugepages information for a pod can be located by an application at path ```/etc/podnetinfo/``` . This directory will contain the request and limit information for 1Gi/2Mb. 

1Gi Hugepages:
* Requests: ```/etc/podnetinfo/hugepages_1G_request_${container-name}```
* Limits:   ```/etc/podnetinfo/hugepages_1G_limit_${container-name}```

2Mi: Hugepages:
* Requests: ```/etc/podnetinfo/hugepages_2M_request_${container-name}```
* Limits:   ```/etc/podnetinfo/hugepages_2M_limit_${container-name}```

pkg/webhook/webhook.go
Billy99: I did change to Warningf - Look at my last commit.

Billy99:
> Can you add to the readme about this multi container support and expected path & labels?

@martinkennelly Good suggestion. Done.

martinkennelly (approved):
LGTM

pperiyasamy:
> I'm still with k8s 1.19.3, will try with K8s 1.20 DownwardAPIHugePages feature gate enabled. thanks.

@Billy99 Yes Bill, This works fine with K8s 1.20.1 after enabling DownwardAPIHugePages  feature gate for api server and kubelet. Thanks for this patch.

PR #58, opened 2021-01-11
Not inject resourceName when it already exists

### LINKED ISSUE #56: Allow control of per-pod injection for a given resource
In a large cluster, user may want to apply a given NRI injection for certain pods, but not the others.
For example, NRI would inject resource requests/limits to first container in the pod, but this doesn't work for pod with multiple containers (with each requesting a userspace device). So user would like to disable resource injection for muiltiple userspace containers and keep using it for the rest of pods.

A possible solution for this issue:

1) Define a k8s label for each NRI supported injection

Such as:
network-resources-injector-downward-api-volume: true
network-resources-injector-k8s-extended-resource: false

These labels will be recognized by NRI and has default value being true.

2) User add the labels in pod spec indicating whether or not a certain resource injection should be enabled/disabled. Default is true (enabled) in the cases that label is not added or label value is not set.
3) Based on the label key and/or value, NRI would enable or disable resource injection for that particular pod.

zshi-redhat:
Fixes #56

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -556,6 +556,17 @@ func createResourcePatch(patch []jsonPatchOperation, Containers []corev1.Contain
 		patch = patchEmptyResources(patch, 0, "limits")
 	}
 
+	for resourceName := range resourceRequests {
+		for _, container := range Containers {
+			if _, exists := container.Resources.Limits[corev1.ResourceName(resourceName)]; exists {
+				delete(resourceRequests, resourceName)
+			}
+			if _, exists := container.Resources.Requests[corev1.ResourceName(resourceName)]; exists {
+				delete(resourceRequests, resourceName)
+			}
+		}
+	}
+
 	resourceList := *getResourceList(resourceRequests)
 
 	for resource, quantity := range resourceList {


pkg/webhook/webhook.go
```diff
@@ -644,17 +661,8 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
...
-					Operation: "add",
-					Path:      "/spec/containers/0/resources/requests/" + toSafeJsonPatchKey(resource.String()), // NOTE: in future we may want to patch specific container (not always the first one)
-					Value:     quantity,
-				})
-				patch = append(patch, jsonPatchOperation{
-					Operation: "add",
-					Path:      "/spec/containers/0/resources/limits/" + toSafeJsonPatchKey(resource.String()),
-					Value:     quantity,
-				})
+			if v, exists := pod.ObjectMeta.Labels[InjectResourceNameLabel]; !exists || v != "false" {
```
pperiyasamy, line 664: doesn't it break when NAD contains `k8s.v1.cni.cncf.io/resourceName` and pod has label with nri.npwg.org/resourcename: false, this would make device not get injected by device plugin and deviceID not set in the subsequent CNI cmdAdd invocation. or should there be a check here not to have both settings together ?

zshi-redhat: Yes, this is intentional. It allows user to skip network resource injection when NRI is running.
The issue it addresses is describe here: https://github.com/k8snetworkplumbingwg/network-resources-injector/issues/56

zshi-redhat:
Rebased against master

pkg/webhook/webhook.go
pperiyasamy: @zshi-redhat ok, but doesn't this make CNI to behave in unexpected way as deviceID is not passed (for example, in ovs-cni attaches veth device into pod container instead of sriov VF, host-device CNI might throw an error which lead pod into container creating error state), are we also fine with not exposing which VF assigned to pod container (atleast via PCI Address env variable from device plugin) ? I'm not sure if I'm missing something here.

zshi-redhat: By skipping the resource injection in NRI, user will need to explicitly define the resource request/limit in pod spec, because NRI is not smart enough to know which container it should inject resource to. 
The default behavior is to always inject resource name as long as it is found in net-attach-def, this is not changed.

pperiyasamy: ok, I understand it now, should we add a check if pod spec contains resource request/limit in any one of the containers when net-attach-def has resource name and resource injection is disabled in pod spec ?

zshi-redhat: I think that's a good idea.
We probably also want to rename the current honorExistingResources flag to something else, because honorExistingResources sounds like it is to disable resource injection when resource is found in pod spec. wdyt?

pperiyasamy: @zshi-redhat now the changes look fine to me by reusing the existing honorExistingResources flag and remove the extra resource requests/limits as directed by networks annotation.

pkg/webhook/webhook.go
```diff
@@ -556,6 +556,17 @@ func createResourcePatch(patch []jsonPatchOperation, Containers []corev1.Contain
 		patch = patchEmptyResources(patch, 0, "limits")
 	}
 
+	for resourceName, _ := range resourceRequests {
```
pperiyasamy, line 559: Just a Nit: golint` for resourceName := range resourceRequests`

pkg/webhook/webhook.go
```diff
@@ -556,6 +556,17 @@ func createResourcePatch(patch []jsonPatchOperation, Containers []corev1.Contain
...
+		for _, container := range Containers {
+			if _, exists := container.Resources.Limits[corev1.ResourceName(resourceName)]; exists {
```
pperiyasamy, line 561: should we also check resource quantities too before deleting entry from resourceRequests map ? when injected resource requests/limits < networks (i.e. pod spec has same network multiple times), in this case should we just adjust the map.entry value ?

zshi-redhat: I think we should either inject or not inject as it's impossible for NRI to predict which container user wants to inject resource to or how many resource per container should be injected if there are multiple containers in pod spec. wdyt?

pkg/webhook/webhook.go
zshi-redhat: Done.

zshi-redhat:
@pperiyasamy any other comments?

pkg/webhook/webhook.go
pperiyasamy: ok, I get it, let's go with disabling all the resource injections when any one of the container's resource request/limit has the same resource requested.

martinkennelly (approved):

PR #59, opened 2021-01-12
Enable user-defined injection for pod annotations

### LINKED ISSUE #57: Support customized injection
Sometimes, user wants to inject customized content in pod manifest on the fly. For example, adding an additional network by injecting net-attach-defs in pod annotation.

This can be achieved by defining a configMap which contains key:value in its data entry, key is a user defined k8s label, value is the content to be injected in pod manifest. NRI inspects the pod label and inject the corresponding data value in pod manifest if the pre-defined label is found. For example:

```
apiVersion: v1
data:
  network-resource-injector-pod-annotation: '{"op": "add", "path": "/metadata/annotations", "value": {"k8s.v1.cni.cncf.io/networks": "sriov-net-attach-def"} }'
kind: ConfigMap
metadata:
  name: network-resource-injector-customized-injection-config
  namespace: network-resource-injector-namespace
```
In above example, `network-resource-injector-pod-annotation` is the k8s label defined by user,  followed by the value (json string) defining customized content to be injected in pod spec.

NRI would listen to the configMap change and record the configmap data entry in its internal data structure. 
During pod creation, NRI receives the pod manifest and inspects the pod labels, if any of pod label is found to match with the key of recorded data entry, NRI would inject the corresponding data entry value to pod manifest.
If configMap is not provided or data is empty in configMap, NRI would skip the customized injection.

zshi-redhat:
User-defined injections allows user to define additinal injections in
Kubernetes ConfigMap and request additional injection for individual
pod based on pod label. Currently user-defined injection only support
injecting pod annotations.

Fixes #57 

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,73 @@ spec:
    master: eno3
 ```
 
+### User Defined Injections
+
+User Defined injections allows user to define additional injections (besides what's supported in NRI, such as ResourceName, Downward API volumes etc) in Kubernetes ConfigMap and request additional injection for individual pod based on pod label. Currently user defined injection only support injecting pod annotations.
+
+In order to use this feature, user needs to create the user defined injection ConfigMap with name `nri-user-defined-injections` in the namespace where NRI was deployed in (`kube-system` namespace is used when there is no `NAMESPACE` environment variable passed to NRI). The data entry in ConfigMap is in the format of key:value pair. Key is a user defined label that will be used to match with pod labels, Value is the actual injection in the format as defined by [RFC6902](https://tools.ietf.org/html/rfc6902) that will be applied to pod manifest. NRI would listen to the creation/update/deletion of this ConfigMap and update its internal data structure every 30 seconds so that subsquential creation of pods will be evaluated against the latest user defined injections.
+
+Metadata.Annotations in Pod definition is the only supported field for customization, whose `path` should be "/metadata/annotations".
+
+Below is an example of user defined injection ConfigMap:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nri-user-defined-injections
+  namespace: kube-system
+data:
+  feature.pod.kubernetes.io/sriov-network: '{"op": "add", "path": "/metadata/annotations", "value": {"k8s.v1.cni.cncf.io/networks": "sriov-net-attach-def"}}'
+```
+
+`feature.pod.kubernetes.io/sriov-network` is a user defined label to request additional networks. Every pod that contains this label with a value set to `"true"` will be applied with the patch that's defined in the following json string.
+
+`'{"op": "add", "path": "/metadata/annotations", "value": {"k8s.v1.cni.cncf.io/networks": "sriov-net -attach-def"}}` defines how/where/what the patch shall be applied.
+
+`"op": "add"` is the action for user defined injection. In above case, it requests to add the `value` to `path` in pod manifests.
+
+`"path": "/metadata/annotations"` is the path in pod manifest to be updated.
+
+`"value": {"k8s.v1.cni.cncf.io/networks": "sriov-net-attach-def"}}` is the value to be updated in the given `path`.
+
+For a pod to request user defined injection, one of its labels shall match with the labels defined in user defined injection ConfigMap.
+For example, with the below pod manifest:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: testpod
+  labels:
+   feature.pod.kubernetes.io/sriov-network: "true"
+spec:
+  containers:
+  - name: app
+    image: centos:7
+    command: [ "/bin/bash", "-c", "--" ]
+    args: [ "while true; do sleep 300000; done;" ]
+```
+
+NRI would update pod manifest to:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: testpod
+  labels:
+   feature.pod.kubernetes.io/sriov-network: "true"
+  annotations:
+    k8s.v1.cni.cncf.io/networks: sriov-net-attach-def
+spec:
+  containers:
+  - name: app
+    image: centos:7
+    command: [ "/bin/bash", "-c", "--" ]
+    args: [ "while true; do sleep 300000; done;" ]
+```
+
 ## Contact Us
 
 For any questions about Network Resources Injector, feel free to ask a question in #network-resources-injector in [NPWG slack](https://npwg-team.slack.com/), or open up a GitHub issue. Request an invite to NPWG slack [here](https://intel-corp.herokuapp.com/).
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -16,19 +16,27 @@ package main
 
 import (
 	"crypto/tls"
+	"context"
 	"flag"
 	"fmt"
 	"net/http"
+	"os"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"github.com/fsnotify/fsnotify"
 	"github.com/golang/glog"
 	"github.com/k8snetworkplumbingwg/network-resources-injector/pkg/webhook"
 )
 
-const defaultClientCa = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+const (
+	defaultClientCa              = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	customizedInjectionConfigMap = "nri-user-defined-injections"
+)
 
 func main() {
+	var namespace string
 	var clientCAPaths webhook.ClientCAFlags
 	/* load configuration */
 	port := flag.Int("port", 8443, "The port on which to serve.")
@@ -54,6 +62,10 @@ func main() {
 		clientCAPaths = append(clientCAPaths, defaultClientCa)
 	}
 
+	if namespace = os.Getenv("NAMESPACE"); namespace == "" {
+		namespace = "kube-system"
+	}
+
 	glog.Infof("starting mutating admission controller for network resources injection")
 
 	keyPair, err := webhook.NewTlsKeypairReloader(*cert, *key)
@@ -67,7 +79,7 @@ func main() {
 	}
 
 	/* init API client */
-	webhook.SetupInClusterClient()
+	clientset := webhook.SetupInClusterClient()
 
 	webhook.SetInjectHugepageDownApi(*injectHugepageDownApi)
 
@@ -169,6 +181,16 @@ func main() {
 				continue
 			}
 			glog.Infof("watcher error: %v", err)
+		case <-time.After(30 * time.Second):
+			cm, err := clientset.CoreV1().ConfigMaps(namespace).Get(
+				context.Background(), customizedInjectionConfigMap, metav1.GetOptions{})
+			if err != nil {
+				if !errors.IsNotFound(err) {
+					glog.Warningf("Failed to get configmap for customized injections: %v", err)
+				}
+				continue
+			}
+			webhook.SetCustomizedInjections(cm)
 		}
 	}
 }
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -32,7 +32,7 @@ import (
 
 const (
 	defaultClientCa              = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-	customizedInjectionConfigMap = "nri-user-defined-injections"
+	userDefinedInjectionConfigMap = "nri-user-defined-injections"
 )
 
 func main() {
@@ -183,10 +183,10 @@ func main() {
 			glog.Infof("watcher error: %v", err)
 		case <-time.After(30 * time.Second):
 			cm, err := clientset.CoreV1().ConfigMaps(namespace).Get(
-				context.Background(), customizedInjectionConfigMap, metav1.GetOptions{})
+				context.Background(), userDefinedInjectionConfigMap, metav1.GetOptions{})
 			if err != nil {
 				if !errors.IsNotFound(err) {
-					glog.Warningf("Failed to get configmap for customized injections: %v", err)
+					glog.Warningf("Failed to get configmap for user-defined injections: %v", err)
 				}
 				continue
 			}
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -20,8 +20,10 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"reflect"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/golang/glog"
 	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
@@ -46,6 +48,11 @@ type jsonPatchOperation struct {
 	Value     interface{} `json:"value,omitempty"`
 }
 
+type customInjections struct {
+	sync.Mutex
+	Patchs map[string]jsonPatchOperation
+}
+
 type hugepageResourceData struct {
 	ResourceName  string
 	ContainerName string
@@ -63,6 +70,7 @@ var (
 	injectHugepageDownApi  bool
 	resourceNameKeys       []string
 	honorExistingResources bool
+	cusInjects             = &customInjections{Patchs: make(map[string]jsonPatchOperation)}
 )
 
 func prepareAdmissionReviewResponse(allowed bool, message string, ar *v1beta1.AdmissionReview) error {
@@ -621,6 +629,51 @@ func getResourceList(resourceRequests map[string]int64) *corev1.ResourceList {
 	return &resourceList
 }
 
+func createCustomizedPatch(pod corev1.Pod) ([]jsonPatchOperation, error) {
+	var customizedPatch []jsonPatchOperation
+
+	// lock for reading
+	cusInjects.Lock()
+	defer cusInjects.Unlock()
+
+	for k, v := range cusInjects.Patchs {
+		// The cusInjects will be injected when:
+		// 1. Pod labels contain the patch key defined in cusInjects, and
+		// 2. The value of patch key in pod labels(not in cusInjects) is "true"
+		if podValue, exists := pod.ObjectMeta.Labels[k]; exists && strings.ToLower(podValue) == "true" {
+			customizedPatch = append(customizedPatch, v)
+		}
+	}
+	return customizedPatch, nil
+}
+
+func getNetworkSelections(annotationKey string, pod corev1.Pod, customizedPatch []jsonPatchOperation) (string, bool) {
+	// User defined annotateKey takes precedence than customized injections
+	glog.Infof("search %s in original pod annotations", annotationKey)
+	nets, exists := pod.ObjectMeta.Annotations[annotationKey]
+	if exists {
+		glog.Infof("%s is defined in original pod annotations", annotationKey)
+		return nets, exists
+	}
+
+	glog.Infof("search %s in customized injections", annotationKey)
+	// customizedPatch may contain user defined net-attach-defs
+	if len(customizedPatch) > 0 {
+		for _, p := range customizedPatch {
+			if p.Operation == "add" && p.Path == "/metadata/annotations" {
+				for k, v := range p.Value.(map[string]interface{}) {
+					if k == annotationKey {
+						glog.Infof("%s is found in customized annotations", annotationKey)
+						return v.(string), true
+					}
+				}
+			}
+		}
+	}
+	glog.Infof("%s is not found in either pod annotations or customized injections", annotationKey)
+	return "", false
+}
+
 // MutateHandler handles AdmissionReview requests and sends responses back to the K8s API server
 func MutateHandler(w http.ResponseWriter, req *http.Request) {
 	glog.Infof("Received mutation request")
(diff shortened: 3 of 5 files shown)


pkg/webhook/webhook.go
```diff
@@ -546,6 +547,68 @@ func createNodeSelectorPatch(patch []jsonPatchOperation, existing map[string]str
...
+	var annotateKey string
+	switch netType {
+	case "default":
+		annotateKey = defaultNetworkAnnotationKey
+	case "additional":
+		annotateKey = networksAnnotationKey
+	}
+
+	// User defined annotateKey takes precedence than customized injections
+	glog.Infof("search %s net-attach-def in original pod annotations", netType)
```
pperiyasamy, line 587: Nit: can't we just log annotateKey instead of netType ? this way networkAnnotationKey can be passed from calling side (instead of netType) which avoids above switch case.

cmd/webhook/main.go
```diff
@@ -16,19 +16,27 @@ package main
...
+	"k8s.io/apimachinery/pkg/api/errors"
 	"github.com/fsnotify/fsnotify"
 	"github.com/golang/glog"
 	"github.com/intel/network-resources-injector/pkg/webhook"
 )
 
-const defaultClientCa = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+const (
+	defaultClientCa              = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	customizedInjectionConfigMap = "nri-customized-injections"
```
pperiyasamy, line 35: would this be only one global config map for all the pods  ?  should we update same existing config map if there is a need to add/update/delete data field ?

pkg/webhook/webhook.go
zshi-redhat: Yes, good catch! updated.

cmd/webhook/main.go
zshi-redhat: Yes, this will be the only global configmap for all pods. A pod can decide which data field shall be applied by using different labels.   This single configmap shall be updated when customized injection will be added/updated/deleted.

pkg/webhook/webhook.go
```diff
@@ -832,3 +894,13 @@ func SetInjectHugepageDownApi(hugepageFlag bool) {
 func SetHonorExistingResources(resourcesHonorFlag bool) {
 	honorExistingResources = resourcesHonorFlag
 }
+
+// SetCustomizedInjections sets additional injections to be applied in Pod spec
+func SetCustomizedInjections(injections *corev1.ConfigMap) {
```
pperiyasamy, line 899: shouldn't we also remove stale entries from customizedInjections Map (to handle delete cases in config map) ?

zshi-redhat: Yes, we should.
Done

pperiyasamy:
/lgtm
@martinkennelly PTAL

pkg/webhook/webhook.go
```diff
@@ -621,6 +622,60 @@ func getResourceList(resourceRequests map[string]int64) *corev1.ResourceList {
...
+		// The customizedInjections will be injected when:
+		// 1. Pod labels contain the patch key defined in customizedInjections, and
+		// 2. The value of patch key in pod labels(not in customizedInjections) is "true"
+		if podValue, exists := pod.ObjectMeta.Labels[cik]; exists && podValue == "true" {
+			err := json.Unmarshal([]byte(civ), &patch)
+			if err != nil {
+				return nil, err
+			}
+		}
+		customizedPatch = append(customizedPatch, patch)
```
martinkennelly, line 639: should "customizedPatch = append(customizedPatch, patch)" be within the previous if statement closure - before ln 638, so we only add new patches?

martinkennelly (commented):
LGTM overall - just one small question and a nit.

pkg/webhook/webhook.go
```diff
@@ -832,3 +894,21 @@ func SetInjectHugepageDownApi(hugepageFlag bool) {
...
+			glog.Infof("Initializing customized injections with key: %v, value: %v", k, v)
+			customizedInjections[k] = v
+		}
+	}
+	// remove stale entries from customized configMap
+	for k, _ := range customizedInjections {
+		if _, ok := injections.Data[k]; ok {
+			continue
+		}
+		glog.Infof("Removing stable entry: %v from customized injections", k)
```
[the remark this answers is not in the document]
zshi-redhat: Done

pkg/webhook/webhook.go
zshi-redhat: Yes, it should.
Done

adrianchiris (commented):
I still need to give this PR a deeper look.

it would be great if:
1. there is some documentation on this feature in README
2. commit message can be more expressive on the what/how

martinkennelly (approved):

zshi-redhat:
> I still need to give this PR a deeper look.
> 
> it would be great if:
> 
> 1. there is some documentation on this feature in README
> 2. commit message can be more expressive on the what/how

Done

README.md
```diff
@@ -179,3 +179,68 @@ spec:
  nodeSelector:
    master: eno3
 ```
+
+### Customized Injections
+
+Customized injections allows user to define additinal injections (besides what's supported in NRI, such as ResourceName, Downward API volumes etc) in Kubernetes ConfigMap and request additional injection for individual pod based on pod label. Currently customized injection only support injecting pod annotations.
+
+In order to use this feature, user needs to create the customized injection ConfigMap with name `nri-customized-injections` in namespace `kube-system`. The data entry in ConfigMap is in the format of key:value pair. Key is a user defined label that will be used to match with pod labels, Value is the actual injection that will be applied to pod manifest. NRI would listen to the creation/update/deletion of this ConfigMap and update its internal data structure every 30 seconds so that subsquential creation of pods will be evaluated against the latest customized injections.
```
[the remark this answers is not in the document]
adrianchiris: missing info here on the format of the injection, there is just an example below but i think a more formal definition is needed.

is it in the format as defined by https://tools.ietf.org/html/rfc6902 ?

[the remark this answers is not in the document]
adrianchiris: In addition you should also state that these user defined patches take precedence over built-in patch operations performed by NRI

README.md
```diff
@@ -179,3 +179,68 @@ spec:
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nri-customized-injections
+  namespace: kube-system
+data:
+  feature.pod.kubernetes.io/sriov-network: '{"op": "add", "path": "/metadata/annotations", "value": {"k8s.v1.cni.cncf.io/networks": "sriov-net-attach-def"}}'
+```
+
+`feature.pod.kubernetes.io/sriov-network` is a user defined label to request additional networks. Every pod that contains this label will be applied with the patch that's defined in the following json string.
```
adrianchiris, line 201: nit: contain this  label with a value set to `true`

pkg/webhook/webhook.go
```diff
@@ -621,6 +622,60 @@ func getResourceList(resourceRequests map[string]int64) *corev1.ResourceList {
...
+		// 2. The value of patch key in pod labels(not in customizedInjections) is "true"
+		if podValue, exists := pod.ObjectMeta.Labels[cik]; exists && podValue == "true" {
+			err := json.Unmarshal([]byte(civ), &patch)
+			if err != nil {
+				return nil, err
+			}
+			customizedPatch = append(customizedPatch, patch)
+		}
+	}
+	for _, p := range customizedPatch {
```
adrianchiris, line 641: can this validation happen on set of `customizedInjections`  instead of every pod creation?

adrianchiris (changes requested):
Thank for proposing this patch Zenghui !

added some comments,

The newly defined methods in webhook.go are fairly isolated, so dare i ask to add unit tests for them ?

pkg/webhook/webhook.go
```diff
@@ -641,8 +696,13 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
...
-	additionalNetSelections, addExists := pod.ObjectMeta.Annotations[networksAnnotationKey]
+	customizedPatch, err := createCustomizedPatch(pod)
+	if err != nil {
+		glog.Warningf("Error, failed to create customized injection patch, %v", err)
+	}
+
+	defaultNetSelection, defExist := getNetworkSelections(defaultNetworkAnnotationKey, pod, customizedPatch)
+	additionalNetSelections, addExists := getNetworkSelections(networksAnnotationKey, pod, customizedPatch)
 
 	if defExist || addExists {
```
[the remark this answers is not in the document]
zshi-redhat: customized patch is evaluated first, even there is no built-in patches are required.

pkg/webhook/webhook.go
```diff
@@ -63,6 +63,7 @@ var (
 	injectHugepageDownApi  bool
 	resourceNameKeys       []string
 	honorExistingResources bool
+	customizedInjections   = make(map[string]string)
```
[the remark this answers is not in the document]
zshi-redhat: could you be specific to the unexpected result? my understanding is if there is a wrong format in customized injection , or incomplete patch or unsupported patch exists in customized Injection, it would cause the createCustomizedPatch fail, the end result would be no customized patch be applied to pod.

cmd/webhook/main.go
```diff
@@ -54,6 +62,10 @@ func main() {
 		clientCAPaths = append(clientCAPaths, defaultClientCa)
 	}
 
+	if namespace = os.Getenv("NAMESPACE"); namespace == "" {
```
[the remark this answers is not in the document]
zshi-redhat: Yes, kube-system is the default namespace, will update server.yam

pkg/webhook/webhook.go
zshi-redhat: `customizedPatch` contains only patch that a pod requests. in the for loop above, there is check on which labels are requested by pod and only add the requested customized patch in `customizedPatch`.
I assume we cannot pre-predict which labels user may set on pod.

README.md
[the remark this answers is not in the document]
zshi-redhat: The order of injection may depend on what customized patch are configured by user. For current annotation, it takes precedence then built-in patch, because there is a chance that annotation may contain net-attach-def definition, that would impact the injeciton of resourceName. I'll add the precedence info.

The format is json string, within which, the value structure is an internal data structure defined at: https://github.com/k8snetworkplumbingwg/network-resources-injector/blob/master/pkg/webhook/webhook.go#L43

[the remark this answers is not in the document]
adrianchiris: > The order of injection may depend on what customized patch are configured by user. For current annotation, it takes precedence then built-in patch, because there is a chance that annotation may contain net-attach-def definition, that would impact the injeciton of resourceName. I'll add the precedence info.

Great, Thanks !


> The format is json string, within which, the value structure is an internal data structure defined at:

So The README section should contain this information IMO (the format) as when a user goes through this section he should understand how to write these custom patches which does match RFC6902

pkg/webhook/webhook.go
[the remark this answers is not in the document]
adrianchiris: > could you be specific to the unexpected result?

`SetCustomizedInjection()` first adds new entries then removes stale entires.
As an example for what might go wrong, a user defines a new config replacing all entries, the map is then updated with new configs added.
Then a pod is created, leading to the injections of both new and old configs and only then the "remove stale entries" part in `SetCustomizedInjection()` takes place.

In general you have a data structure that is written and read asynchronously in a non atomic manner by different goroutines, it is generally best practice to add some sort of synchronization mechanism to avoid any inconsistencies when reading the data structure.

pkg/webhook/webhook.go
adrianchiris: it is necessarily a subset of `customizedInjections ` isnt it ?
so validating that each patch provided in `customizedInjections` once when setting it will not require to check it per Pod.
it seems to me more efficient.

This does not affect the correctness of the code, and is a suggestion on how to make it a bit more efficient as less operations will be performed when processing a Pod.

pkg/webhook/webhook.go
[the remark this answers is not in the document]
adrianchiris: In the code i see that customized patches are applied only if default network annotation exists or secondary network annotation exists in either the original pod annotation OR as a result of a user-defined labels with a customized injection matching these labels.

Is that the intended behaviour ?
Or should the customized Patch be applied to the pod regardless ?

README.md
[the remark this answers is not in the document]
zshi-redhat: Added RFC6902 reference link.

README.md
```diff
@@ -179,3 +179,68 @@ spec:
  nodeSelector:
    master: eno3
 ```
+
+### Customized Injections
```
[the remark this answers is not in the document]
zshi-redhat: Updated to `User Defined Injections` for this title.

cmd/webhook/main.go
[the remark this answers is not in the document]
zshi-redhat: Done.

pkg/webhook/webhook.go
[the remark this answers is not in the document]
zshi-redhat: Added a read/write lock for global structure data update.

pkg/webhook/webhook.go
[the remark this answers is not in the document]
zshi-redhat: This is intended behaviour, at least for now. We want to inject the patchs only for pod requesting resources.

README.md
zshi-redhat: Done.

pkg/webhook/webhook.go
zshi-redhat: Moved this down to `SetCustomizedInjection()` so that it only be executed once.

zshi-redhat:
> 
> The newly defined methods in webhook.go are fairly isolated, so dare i ask to add unit tests for them ?

unit tests added in second commit.

README.md
```diff
@@ -179,3 +179,68 @@ spec:
  nodeSelector:
    master: eno3
 ```
+
+### User Defined Injections
+
+User Defined injections allows user to define additinal injections (besides what's supported in NRI, such as ResourceName, Downward API volumes etc) in Kubernetes ConfigMap and request additional injection for individual pod based on pod label. Currently customized injection only support injecting pod annotations.
```
martinkennelly, line 185: nit: additinal -> additional

README.md
```diff
@@ -179,3 +179,68 @@ spec:
...
+   feature.pod.kubernetes.io/sriov-network: "true"
+  annotations:
+    k8s.v1.cni.cncf.io/networks: sriov-net-attach-def
+spec:
+  containers:
+  - name: app
+    image: centos:7
+    command: [ "/bin/bash", "-c", "--" ]
+    args: [ "while true; do sleep 300000; done;" ]
+```
```
martinkennelly, line 246: awesome description - thanks!

pkg/webhook/webhook.go
```diff
@@ -621,6 +629,51 @@ func getResourceList(resourceRequests map[string]int64) *corev1.ResourceList {
...
+
+	// lock for reading
+	cusInjects.Lock()
+	defer cusInjects.Unlock()
+
+	for k, v := range cusInjects.Patchs {
+		// The cusInjects will be injected when:
+		// 1. Pod labels contain the patch key defined in cusInjects, and
+		// 2. The value of patch key in pod labels(not in cusInjects) is "true"
+		if podValue, exists := pod.ObjectMeta.Labels[k]; exists && podValue == "true" {
```
martinkennelly, line 643: do you think its worthwhile to cast podValue to lowercase characters before comparison with true?

zshi-redhat: Yes, it'd be better to use lowercase.
Updated.

README.md
zshi-redhat: Good catch!
Done.

martinkennelly (approved):

zshi-redhat:
@adrianchiris PTAL on the latest changes.

README.md
```diff
@@ -180,6 +180,71 @@ spec:
    master: eno3
 ```
 
+### User Defined Injections
+
+User Defined injections allows user to define additional injections (besides what's supported in NRI, such as ResourceName, Downward API volumes etc) in Kubernetes ConfigMap and request additional injection for individual pod based on pod label. Currently customized injection only support injecting pod annotations.
```
adrianchiris, line 185: IMO if user defined injections is used, we should use it all the way.
in the content of this section, examples and code. The word "customized" should be replaced with "user defined" so we are consistent throughout.

pkg/webhook/webhook.go
```diff
@@ -832,3 +892,43 @@ func SetInjectHugepageDownApi(hugepageFlag bool) {
...
+		// unmarshall customized injection to json patch
+		err := json.Unmarshal([]byte(v), &patch)
+		if err != nil {
+			glog.Errorf("Failed to unmarshall customized injection: %v", v)
+			continue
+		}
+		// metadata.Annotations is the only supported field for customization
+		// jsonPatchOperation.Path should be "/metadata/annotations" when customizing pod annotation
+		if patch.Path != "/metadata/annotations" {
+			glog.Errorf("Path: %v is not supported, only /metadata/annotations can be customized", patch.Path)
```
adrianchiris, line 912: The limitation we impose on the path should be mentioned in the README

pkg/webhook/webhook.go
```diff
@@ -832,3 +892,43 @@ func SetInjectHugepageDownApi(hugepageFlag bool) {
...
+	for k, _ := range cusPatchs {
+		if _, ok := injections.Data[k]; ok {
+			continue
+		}
+		glog.Infof("Removing stale entry: %v from customized injections", k)
+		delete(cusPatchs, k)
+	}
+
+	// lock for writing
+	cusInjects.Lock()
```
adrianchiris, line 930: you are modifying the map at both L#917 and L#926 as `cusPatchs`  is a reference to the original map

both add and delete operations are not atomic to my knowledge so you should either lock/unlock at the critical points or lock at the beginning and defer unlock.

pkg/webhook/webhook.go
```diff
@@ -832,3 +892,43 @@ func SetInjectHugepageDownApi(hugepageFlag bool) {
...
+	defer cusInjects.Unlock()
+	cusInjects.Patchs = cusPatchs
```
adrianchiris, line 932: no need to do this assignment.

https://play.golang.org/p/pHs2-7u6VXW

adrianchiris (changes requested):

pkg/webhook/webhook.go
zshi-redhat: Done

pkg/webhook/webhook.go
zshi-redhat: Done

pkg/webhook/webhook.go
zshi-redhat: Done

pkg/webhook/webhook_test.go
```diff
@@ -127,6 +128,235 @@ var _ = Describe("Webhook", func() {
...
+			map[string]jsonPatchOperation{
+				"nri-inject-annotation": jsonPatchOperation{
+					Operation: "add",
+					Path: "/metadata/annotations",
+					Value: map[string]interface{}{"v1.multus-cni.io/default-network": "sriov-net-new"},
+				},
+			},
+		),
+	)
+
```
adrianchiris, line 359: Could you add a test that ensures a patch operation fails if the Path is not prefixed with "/metadata/" ?
we should ensure this does not change unintentionally in the future.

cmd/webhook/main.go
```diff
@@ -16,19 +16,27 @@ package main
...
+	"k8s.io/apimachinery/pkg/api/errors"
 	"github.com/fsnotify/fsnotify"
 	"github.com/golang/glog"
 	"github.com/k8snetworkplumbingwg/network-resources-injector/pkg/webhook"
 )
 
-const defaultClientCa = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+const (
+	defaultClientCa              = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	customizedInjectionConfigMap = "nri-user-defined-injections"
```
adrianchiris, line 35: it would be nice if we didnt have the word "customized" in code since we are referring to this as "user-defined"

e.g `customizedInjectionConfigMap` -> `userDefinedInjectionConfigMap` and so on.

not something i would block on, your call if you want to do the extra refactor :)

adrianchiris (commented):
Thanks for addressing my comments.

small nit on the unit-tests, the other comment is just a suggestion (optional)

cmd/webhook/main.go
zshi-redhat: Done, it's easy to update the name than the code logic :-)

pkg/webhook/webhook_test.go
zshi-redhat: This unit test would fail if other path is used:
https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/59/files#diff-54553ccca2e329f7bf1fdf04d002ced7323d3fa477e7fc6e2a0a9d248cb10ab2R313-R322
do you think it is enough?

adrianchiris: Ah, missed this one !

adrianchiris (approved):
Thanks for addressing my comments !

adrianchiris:
@zshi-redhat mind rebasing this ? once we are green this can be merged IMO

zshi-redhat:
> @zshi-redhat mind rebasing this ? once we are green this can be merged IMO

Rebased

Issue #48, opened 2020-12-14
NRI doesn't catch termination signals and end


martinkennelly:
### What happened?
NRI doesn't catch and end cleanly when termination signals are received.
### What did you expect to happen?
NRI catches ```SIGTERM```, ```SIGHUP```, ````SIGINT```, ```SIGQUIT``` and  ends.

martinkennelly:
I am currently working on fixing this issue.

PR #62, opened 2021-01-19
Catch termination signal

martinkennelly:
In order for termination signal to be captured, two goroutines
are created which handle TLS keypair reloading and HTTP server
instantiation.

Signed-off-by: Kennelly, Martin <martin.kennelly@intel.com>

Fixes: #48


martinkennelly:
I want to update this patch to make it easier to test.
I will add test cases.

martinkennelly:
A lot of changes here I know. I improved the test coverage to 33%. I am hoping to add e2e tests in a separate PR using kind and this will help validate this patch.

martinkennelly:
@zshi-redhat Do you still see value with this patch? I was going to push for it once E2E test cases got to a sufficient level where we can do regression testing of the features. @MichalGuzieniuk is working on adding these right now.

I want this patch for the following reasons but it will need to be reworked to accommodate your user defined injections work but that is no biggie once we have tests for it.

Reasons:
* NRI will catch term signals - currently it doesn't handle them.
* NRI will end if HTTP server fails - currently nothing happens if it fails.
* NRI will end if keypair reloader fails - current we just keep retrying.
* Improved unit test coverage from ~20% to 33%
* Improved separation of code instead of it all bundled into webhook.pkg. We need to stop that going forward imo.

I propose adding user injections golang routine as well and if it fails, NRI should end, and be re-provisioned.

Anyway, Its up to you if you see any value with this patch long term. Let me know when you have time. No rush as I think its good to fix the existing problem with user injection first, and increase the e2e test cases (thanks to @MichalGuzieniuk ) to ensure we have regression testing when this major change occurs.

zshi-redhat:
> @zshi-redhat Do you see value with this patch? I was going to push for it once E2E test cases got to a sufficient level where we can do regression testing of the features. @MichalGuzieniuk is working on adding these right now.

+1, yes, absolutely.
> 
> I want this patch for the following reasons but it will need to be reworked to accommodate your user defined injections work but that is no biggie once we have tests for it.
> 
> Reasons:
> 
>     * NRI will catch term signals - currently it doesn't handle them.
> 
>     * NRI will end if HTTP server fails - currently nothing happens if it fails.
> 
>     * NRI will end if keypair reloader fails - current we just keep retrying.
> 
>     * Improved unit test coverage from ~20% to 33%
> 
>     * Improved separation of code instead of it all bundled into webhook.go. We need to stop that going forward imo.
> 
I think these are positive moves.
I had the same idea of separating giant webhook.go into multiple files, didn't get chance to implement it :-) 

> 
> I propose adding user injections golang routine as well and if it fails, NRI should end, and be re-provisioned.

User-defined-injections is an optional feature for now, We probably don't want to end the NRI process for the absense of configMap or so.

> 
> Anyway, Its up to you if you see any value with this patch long term. Let me know when you have time. No rush as I think its good to fix the existing problem with user injection first, and increase the e2e test cases (thanks to @MichalGuzieniuk ) to ensure we have regression testing when this major change occurs.

Thanks for working on these improvements!

pkg/webhook/channel.go
```diff
@@ -0,0 +1,92 @@
...
+func (c *Channel) WaitUntilClosed(limit time.Duration) error {
+	if interval > limit {
+		return errors.New("limit arg value too low")
+	}
+	tEnd := time.Now().Add(limit)
+	for tEnd.After(time.Now()) {
+		if c.IsClosed() {
+			return nil
+		}
+		time.Sleep(interval)
```
pperiyasamy, line 74: can we derive sleep interval from limit ?

pkg/webhook/channel.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+)
+
+const (
+	interval = time.Millisecond * 2
+	chBuffer = 1
```
pperiyasamy, line 12: can move buffer size into Channel and configurable from NewChannel API?

pkg/webhook/channel.go
```diff
@@ -0,0 +1,92 @@
+package webhook
```
pperiyasamy, line 1: can we move this into another package pkg/tools/channel.go ?

pkg/webhook/webhook.go
```diff
@@ -832,3 +835,44 @@ func SetInjectHugepageDownApi(hugepageFlag bool) {
 func SetHonorExistingResources(resourcesHonorFlag bool) {
 	honorExistingResources = resourcesHonorFlag
 }
+
+//Watch blocks until either TLS cert & key watcher or HTTP server or termination signal generated
+func Watch(server nri.Service, watcher nri.Service, term chan os.Signal) (err error) {
```
pperiyasamy, line 840: can we make this method more generic by making nri.Service as var-args ? 
this would also enable nad cache service for handling termination signal.

pkg/webhook/webhook.go
```diff
@@ -832,3 +835,44 @@ func SetInjectHugepageDownApi(hugepageFlag bool) {
...
+	signal.Notify(term, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	select {
+	case <-watcher.StatusSignal(): // when TLS cert & key watcher finishes unexpectedly
+		glog.Error("TLS key & cert watcher ended")
+		err = server.Quit()
+	case <-server.StatusSignal(): // when HTTP server finishes unexpectedly
```
pperiyasamy, line 846: a question: when write happens on this server channel ? i couldn't find it anywhere  in the code :)

zshi-redhat:
@martinkennelly Would you mind rebase this PR? Thank you!

pkg/webhook/channel.go
martinkennelly: I think in golang, its an antipattern to put funcs in packages named "tools" or "utils". 
["Name your packages after what they provide, not what they contain."](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common)
However, having everything in package webhook is not good either and something I want to improve. I will move the channel stuff into its package. Thanks @pperiyasamy

pkg/webhook/webhook.go
martinkennelly: @pperiyasamy Good idea but I don't know how I could use select to watch all the service signals. If you can think of a pattern let me know.

pkg/webhook/webhook.go
martinkennelly: @pperiyasamy see server.go: within method with pointer receiver ```Run()``` -> ```defer mSrv.Status.Close()```. When the channel closes: 

> After calling close, and after any previously sent values have been received, receive operations will return the zero value for the channel's type without blocking.
 

ref: https://golang.org/ref/spec#Close

.. this causes this case to trigger

pkg/webhook/channel.go
pperiyasamy: good info! I will follow this convention ;)

pkg/webhook/webhook.go
pperiyasamy: ah! ok, that means server.go's` mSrv.instance.Start() `is blocking call, right ?

pkg/webhook/webhook.go
pperiyasamy: can we try something like this ?

```
	ch := make(chan int)
	for idx := range services {
		go func(serv Service, servIdx int) {
			<-serv.StatusSignal()
			for _, otherService := range services[servIdx+1:] {
				otherService.Quit()
			}
			close(ch)
		}(services[idx], idx)
	}
	<-ch
        return
```

martinkennelly:
Closing this as I don't have time to work on it. Anyone feel free to carry it on.

MichalGuzieniuk:
@martinkennelly if you don't mind I will take codebase done by you in this PR, make the rebase and I will continue work on it.

martinkennelly:
@martinkennelly Don't mind at all. I got stuck continuously refactoring NRI with this PR. Here is the outcome: https://github.com/martinkennelly/network-resources-injector/commits/refactor_webhook

The refactor went too far and I was concerned with that... I hope you can do a better job limiting the refactor scope than I did..
I think that refactor is good long term, just the scope of changes...

martinkennelly:
I converted everything into services which could be watched, killed.  I separated out all the code into pkgs as well.

PR #64, opened 2021-02-16
Add build docker image GitHub action

martinkennelly:
For each new PR, we want to build the Dockerfile using GitHub actions.

I attempted to add the dockerfile.rhel7 but the build failed.
Part of issue #68

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,16 +1,18 @@
 name: Docker-build
-on: [pull_request]
+on: [pull_request, push]
 jobs:
-
   build:
-    name: Docker build
     runs-on: ubuntu-latest
     env:
       GO111MODULE: on
-      TAG: network-resources-injector:latest 
     steps:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
-    - name: Build Dockerfile
-      run: docker build -f Dockerfile -t ${TAG} .
+    - name: Build only
+      uses: docker/build-push-action@v2
+      with:
+        push: false
+        context: .
+        file: ./Dockerfile
+        tags: network-resources-injector:latest
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,16 @@
+name: Docker-build
+on: [pull_request]
+jobs:
+
+  build:
+    name: Docker build
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+      TAG: network-resources-injector:latest 
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Build Dockerfile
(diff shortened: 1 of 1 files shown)


martinkennelly:
I you can also add cron job to these actions - do we want to add this to this PR? Twice a week ?

zshi-redhat:
> For each new PR, we want to build the Dockerfile using GitHub actions.
> 
> I attempted to add the dockerfile.rhel7 but the build failed.
> Part of issue #68

I think we might want to just remove the Dockerfile.rhel7 as it is for openshift.

zshi-redhat:
> you can also add cron job to these actions - do we want to add this to this PR? Twice a week ?

If we run test on every PR, would it be already sufficient?

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+      TAG: network-resources-injector:latest 
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Build Dockerfile
+      run: docker build -f Dockerfile -t ${TAG} .
```
zshi-redhat, line 16: nit: I see there is a docker build-and-push github action, do you know if we can use it for build image only?

zshi-redhat:
btw, will this workflow also aim to push the docker image to github container registry?

martinkennelly:
> > you can also add cron job to these actions - do we want to add this to this PR? Twice a week ?
> 
> If we run test on every PR, would it be already sufficient?
I'd like to run it on a regular time cadence as well in case any dependencies that aren't locked to a version change and break NRI.

.github/workflows/docker-build.yml
martinkennelly: > btw, will this workflow also aim to push the docker image to github container registry?

I added it to latest commit with build image only! Thanks for this! I also see there is support for multi arch docker builds. I attempted it but it [hung](https://github.com/martinkennelly/network-resources-injector/actions/runs/632419982).

martinkennelly:
> btw, will this workflow also aim to push the docker image to github container registry?

I think it would be better to have this in a separate workflow. The reason is we want a workflow for testing PRs / pushes but we don't want to build an image for each PR / pushes. I think having a separate workflow to accomplish github container pushes is a good idea which will push when we merge anything to master and also for new releases.

zshi-redhat:
@martinkennelly Do we still need this PR given https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/83 is merged?

martinkennelly:
@zshi-redhat I think #83 should not hinder this PR because it only builds images for tagged releases or merges to master. I think a case can definitely be made for  #72 (add e2e tests with KinD) because it will build the change and then run e2e tests. It doesn't seem necessary to also build again with this workflow. I will close because of this. LMK if you disagree.

PR #69, opened 2021-02-17
Remove buildmode=pie

manuelbuil:
By using buildmode=pie dynamic linking is activated and as a
consequence, the binary depends on /lib/ld-musl-x86_64.so.1,
which makes it work only in Alpine

Besides, this is at odds with CGO_ENABLED=0 option which is
part of the script

Signed-off-by: Manuel Buil <mbuil@suse.com>

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -29,5 +29,5 @@ export GOBIN=${PWD}/bin
 export CGO_ENABLED=0
 export GO15VENDOREXPERIMENT=1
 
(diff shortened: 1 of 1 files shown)


martinkennelly (approved):
Hey,
I added this in after following internal processes and it doesn't make sense after reading about both of these options for golang and CGO_ENABLED disabled. I didn't understand this correctly when I added it. Thanks for raising this.

zshi-redhat (approved):

PR #70, opened 2021-02-17
Remove installation of packages

manuelbuil:
The Dockerfile installs some packages which are not needed to build the
project

Signed-off-by: Manuel Buil <mbuil@suse.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@
 FROM golang:1.13-alpine as builder
 COPY . /usr/src/network-resources-injector
 WORKDIR /usr/src/network-resources-injector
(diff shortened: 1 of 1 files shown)


martinkennelly (commented):
This doesn't work on my system. Make is missing dep.

manuelbuil:
> This doesn't work on my system. Make is missing dep.

thanks for checking! I made a mistake, we should just remove the linux-headers since I don't think they are needed

martinkennelly (approved):
LGTM - Accomplished basic testing and it works fine. Thanks @manuelbuil

zshi-redhat (approved):

PR #77, opened 2021-03-09
Update gogo/protobuf to v1.3.2

zshi-redhat:
This commit updates protobuf to v1.3.2 to
address CVE-2021-3121

Steps to update:
go get github.com/gogo/protobuf@v1.3.2

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,11 +11,13 @@ require (
 	github.com/onsi/gomega v1.10.1
 	github.com/pkg/errors v0.9.1
 	golang.org/x/crypto v0.0.0-20201221181555-eec23a3978ad // indirect
-	golang.org/x/text v0.3.3 // indirect
 	gopkg.in/intel/multus-cni.v3 v3.4.2
 	k8s.io/api v0.18.5
 	k8s.io/apimachinery v0.18.5
 	k8s.io/client-go v11.0.0+incompatible
 )
 
-replace k8s.io/client-go => k8s.io/client-go v0.18.5
+replace (
+	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
+	k8s.io/client-go => k8s.io/client-go v0.18.5
+)
--- a/go.sum
+++ b/go.sum
@@ -74,10 +74,8 @@ github.com/go-openapi/swag v0.19.2/go.mod h1:POnQmlKehdgb5mhVOsnJFsivZCEZ/vjK9gh
 github.com/go-openapi/swag v0.19.5/go.mod h1:POnQmlKehdgb5mhVOsnJFsivZCEZ/vjK9gh66Z9tfKk=
 github.com/go-sql-driver/mysql v1.3.0/go.mod h1:zAC/RDZ24gD3HViQzih4MyKcchzm+sOG5ZlKdlhCg5w=
 github.com/godbus/dbus v0.0.0-20180201030542-885f9cc04c9c/go.mod h1:/YcGZj5zSblfDWMMoOzV4fas9FZnQYTkDnsGvmh2Grw=
-github.com/gogo/protobuf v1.1.1/go.mod h1:r8qH/GZQm5c6nD/R0oafs1akxWv10x8SbQlK7atdtwQ=
-github.com/gogo/protobuf v1.3.0/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
-github.com/gogo/protobuf v1.3.1 h1:DqDEcV5aeaTmdFBePNpYsp3FlcVH/2ISVVM9Qf8PSls=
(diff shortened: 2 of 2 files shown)


zshi-redhat:
perhaps we don't need this fix as gogo/protobuf is not used in network-resources-injector, even we have old urls are present in go.sum.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/cloudflare/cfssl v1.4.1
 	github.com/fsnotify/fsnotify v1.4.9
+	github.com/gogo/protobuf v1.3.2 // indirect
```
adrianchiris, line 8: should this be put in a replace directive ? 

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/015d6e055bf0bc27f6d8da74509b3bee433e9613/go.mod#L44

zshi-redhat: I will make the change.
So if we have multiple versions in go.sum, it means all those versions are used directly/indirectly in the repo, not just the tip version?

adrianchiris: Thats what i understand from: https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive

adrianchiris: looking at `go.sum` , i see gogo/protobuf v1.3.0 and 1.3.1 which is not present if you use the replace statement.
see : https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/015d6e055bf0bc27f6d8da74509b3bee433e9613/go.sum#L310

zshi-redhat: > looking at `go.sum` , i see gogo/protobuf v1.3.0 and 1.3.1 which is not present if you use the replace statement.
> see : https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/015d6e055bf0bc27f6d8da74509b3bee433e9613/go.sum#L310
> 
> if you delete and recreate go.sum file are these old versions still there ? I assume if they are there, something is using it.

They are still there if I re-create go.sum, but they may not be used: https://github.com/golang/go/wiki/Modules#is-gosum-a-lock-file-why-does-gosum-include-information-for-module-versions-i-am-no-longer-using

martinkennelly: Minimum selection algorithm ensures 1.3.1 is chosen for version 1 of protobuf. See [here](https://github.com/golang/go/wiki/Modules#version-selection). Its fine if lower versions of v1 is seen in go.sum. It will not be chosen.
Going forward, are we using go get and verify with go list -m all? or replace directive? I'd rather go get but I have no reason why except its what I see others use. We should also run go mod tidy after any action on go.mod.

martinkennelly (approved):

martinkennelly (changes requested):
could you run go mod tidy as well. Thank you.

zshi-redhat:
> could you run go mod tidy as well. Thank you.

Done. Thanks Martin for the review!

go.mod
martinkennelly: @adrianchiris are you ok with this path - using "go get ..." instead of replace directive to force upgrading components? Or would you rather use replace directive? I am open to hearing your experienced experience on this and knowledge from how other repos do it. I think go get and verify with go list -m all is fine for upgrading.

martinkennelly (approved):
Approved. Don't see any reason why I should gate this. Id like us to agree on an upgrade mechanism. It seems replace directive is now agreed as the way to do it. Thanks Zenghui/Adrian.

zshi-redhat:
> Approved. Don't see any reason why I should gate this. Id like us to agree on an upgrade mechanism. It seems replace directive is now agreed as the way to do it. Thanks Zenghui/Adrian.

@martinkennelly I was pointed to how ovn-k8s did the update by Adrian: https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/go.mod#L47

I didn't investigate to the bottom whether old links in go.sum are still used in building images which Adrian believe might be the case. Adrian suggested to try build a image and see if those old versions are downloaded in container, If that's the case, those old links might still needed.

martinkennelly:
@zshi-redhat I see. I'm fine with using the replace directive. Good to have an agreed mechanism.

martinkennelly:
@zshi-redhat saw your commit message, you can add replace with this command:
```
go mod edit -replace github.com/repo@lower_ver=github.com/repo@higher_ver 
```

go.mod
adrianchiris: > Minimum selection algorithm ensures 1.3.2 is chosen for version 1 of protobuf. See here. Its fine if lower versions of v1 is seen in go.sum. It will not be chosen.
Going forward, are we using go get and verify with go list -m all? or replace directive? I'd rather go get but I have no reason why except its what I see others use. We should also run go mod tidy after any action on go.mod.

I see, thanks for looking at this Martin!
So it seems that both methods: go get , replace are valid in overriding old module version.
i know the latter works. And asked zenghui to ensure the former indeed works before we merge.

So, TL;DR, if both works i prefer the "go get" as if later on we depend on a module with a newer version of a dependency, we will not break the module by overriding the dependency with a lower version.

PR #79, opened 2021-03-15
add nod selector though resource requests empty

### LINKED ISSUE #78: node selector injection feature doesn't work when ResourceName is not defined in NAD
Hey,
During e2e testing, we noticed that node selector injection doesn't work when ```resourceName``` is not defined in the NAD.
Works fine when there is an additional network defined.
I believe a use case was when no additional network was defined. i.e. [macvlan](https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/21#issuecomment-597936699)
Documentation example for readme: #50 
Original PR: #21 
@pperiyasamy 

Node selector injection feature works:
```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  name: foo-network
  annotations:
    k8s.v1.cni.cncf.io/nodeSelector: kubernetes.io/hostname=kind-worker2
    k8s.v1.cni.cncf.io/resourceName: example.com/foo
spec:
...
```

Node selector injection feature doesnt work:
```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  name: foo-network
  annotations:
    k8s.v1.cni.cncf.io/nodeSelector: kubernetes.io/hostname=kind-worker2
spec:
...
```

Sample Macvlan NAD with no ```resourceName``` defined:
```
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: macvlan-conf
spec: 
  config: '{
      "cniVersion": "0.3.1",
      "plugins": [
        {
          "type": "macvlan",
          "capabilities": { "ips": true },
          "master": "eth0",
          "mode": "bridge",
          "ipam": {
            "type": "static",
            "routes": [
              {
                "dst": "0.0.0.0/0",
                "gw": "10.1.1.1"
              }
            ] 
          }
        }, {
          "capabilities": { "mac": true },
          "type": "tuning"
        }
      ]
    }'
```

pperiyasamy:
The NRI currently injects node selector in case of resource requests are not empty. But there are cases when pod would request secondary networks without a need for resource name specified in the NAD (example: macvlan network).  
Hence this PR attempts to fix it.

Fixes #78 

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -758,11 +758,10 @@ func MutateHandler(w http.ResponseWriter, req *http.Request) {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
-
+		var patch []jsonPatchOperation
 		if len(resourceRequests) == 0 {
 			glog.Infof("pod doesn't need any custom network resources")
 		} else {
-			var patch []jsonPatchOperation
 			glog.Infof("honor-resources=%v", honorExistingResources)
 			if honorExistingResources {
 				patch = updateResourcePatch(patch, pod.Spec.Containers, resourceRequests)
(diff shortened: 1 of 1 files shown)


JanScheurich:
I have some doubts that NRI should unconditionally inject all these things into each pod that uses secondary networks.
For NADs that contain a node-selector annotation, the node selector of course should be injected into the pod, i.e. the actual bug #78 must be fixed.

The podnetinfo network-status annotation may be useful for pods with secondary networks, but it is less obvious that all pods would want this mounted. It could also easily conflict with existing mounts of the pod/containers. I'd rather not do this unconditionally. For NADs with device resources, it should be mounted, though.

Hugepage information is typically only relevant for DPDK applications and should probably only be injected if the NAD refers device resources that are to be or can be consumed as DPDK interfaces. Is that visible from the NAD annotation?

pperiyasamy:
@JanScheurich  Thanks for the clarification, updated the PR only to inject node selector.

martinkennelly (approved):

zshi-redhat (approved):

PR #80, opened 2021-03-16
Add constraints to pod spec

martinkennelly:
Small changes to pod spec to improve security
See commits.

I wanted to group ```allowPrivilegeEscalation``` & ```readOnlyRootFilesystem``` at the "pod level" ```securityContext``` (ln 23) but it will not allow me.

### DIFF
--- a/deployments/server.yaml
+++ b/deployments/server.yaml
@@ -75,6 +75,13 @@ spec:
     volumeMounts:
     - name: tls
       mountPath: /etc/tls
+    resources:
+      requests:
+        memory: "50Mi"
+        cpu: "250m"
+      limits:
+        memory: "200Mi"
+        cpu: "500m"
   volumes:
   - name: tls
     emptyDir: {}
--- a/deployments/server.yaml
+++ b/deployments/server.yaml
@@ -72,6 +72,9 @@ spec:
     securityContext:
       allowPrivilegeEscalation: false
       readOnlyRootFilesystem: true
+      capabilities:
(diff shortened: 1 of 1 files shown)


deployments/server.yaml
```diff
@@ -20,6 +20,10 @@ metadata:
   name: network-resources-injector
   namespace: kube-system
 spec:
+  securityContext:
+    runAsUser: 10000
+    runAsGroup: 10000
+    runAsNonRoot: true
```
zshi-redhat, line 26: Does runAsNonRoot require setting nonRoot user in Dockerfile?

martinkennelly: @zshi-redhat Unfortunately, it seems that it does for NRI. I tried to remove both of the runAsUser and runAsGroup and I got the following error:
```
Error: container has runAsNonRoot and image will run as root
```

zshi-redhat: @martinkennelly IIUC, we need to update Dockerfile to use non-root user, right?
If yes, would you mind updating the Dockerfile in this PR?

martinkennelly: @zshi-redhat Good idea. I tried it but k8 API doesn't even allow it pass through for scheduling. See the runAsNonRoot validate here: https://github.com/kubernetes/kubernetes/blob/5648200571889140ad246feb82c8f80a5946f167/pkg/kubelet/kuberuntime/security_context.go#L91

Issue #84, opened 2021-04-07
User Defined Injections - documentation is misleading


MichalGuzieniuk:
The ConfigMap given as an example at https://github.com/k8snetworkplumbingwg/network-resources-injector#user-defined-injections is wrong, and in result it is not possible to take example as it is and verify if User Defined Injections feature of NRI is working correctly.

ConfigMap does not support '/' character. There is an issue described at https://github.com/kubernetes/kubernetes/issues/87119. In result the key of data section **feature.pod.kubernetes.io/sriov-network:** is invalid and not accepted by kubernetes.

It is described within documentation https://pkg.go.dev/k8s.io/api/core/v1#ConfigMap that 

> Each key must consist of alphanumeric characters, '-', '_' or '.'.

PR #97, opened 2021-04-26
Update documentation regarding user defined injections

MichalGuzieniuk:
This pull requests fixes https://github.com/k8snetworkplumbingwg/network-resources-injector/issues/84

It updates documentation regarding user defined injection. Now it will be possible to take examples as it is and apply it to cluster without need to do modifications. I also add notes for the users that shows current limitations.

In addition, I added table of content.


zshi-redhat:
@MichalGuzieniuk  thanks for fixing the doc and adding the notes!

/lgtm

zshi-redhat (approved):

PR #100, opened 2021-05-18
Ignore AdmissionReview request with empty namespace

zshi-redhat:
https://bugzilla.redhat.com/show_bug.cgi?id=1959520

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -154,6 +154,13 @@ func deserializePod(ar *v1beta1.AdmissionReview) (corev1.Pod, error) {
 	if pod.ObjectMeta.Namespace != "" {
 		return pod, err
 	}
+
+	// AdmissionRequest contains an optional Namespace field
+	if ar.Request.Namespace != "" {
+		pod.ObjectMeta.Namespace = ar.Request.Namespace
+		return pod, nil
+	}
+
 	ownerRef := pod.ObjectMeta.OwnerReferences
 	if ownerRef != nil && len(ownerRef) > 0 {
 		namespace, err := getNamespaceFromOwnerReference(pod.ObjectMeta.OwnerReferences[0])
@@ -162,6 +169,10 @@ func deserializePod(ar *v1beta1.AdmissionReview) (corev1.Pod, error) {
 		}
 		pod.ObjectMeta.Namespace = namespace
 	}
+
+	// pod.ObjectMeta.Namespace may still be empty at this point,
+	// but there is a chance that net-attach-def annotation contains
+	// a valid namespace
 	return pod, err
 }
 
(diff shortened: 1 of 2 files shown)


pkg/webhook/webhook.go
```diff
@@ -270,7 +281,18 @@ func parsePodNetworkSelections(podNetworks, defaultNamespace string) ([]*multus.
...
-			networkSelection.Namespace = defaultNamespace
+			if defaultNamespace == "" {
+				// Ignore the AdmissionReview request when the following conditions are met:
+				// 1) net-attach-def annotation doesn't contain a valid namespace
+				// 2) defaultNamespace retrieved from admission request is empty
+				// Pod admission would fail in subsquent call "getNetworkAttachmentDefinition"
+				// if no namespace is specified. We don't want to fail the pod creation
+				// in such case since it is possible that pod is not a SR-IOV pod
+				glog.Warningf("The admission request doesn't contain a valid namespace, ignoring...")
+				return nil, nil
```
pperiyasamy, line 292: shouldn't you return empty NetworkSelectionElement array here instead of nil ?

pkg/webhook/webhook.go
```diff
@@ -154,6 +154,13 @@ func deserializePod(ar *v1beta1.AdmissionReview) (corev1.Pod, error) {
 	if pod.ObjectMeta.Namespace != "" {
 		return pod, err
 	}
+
+	// AdmissionRequest contains an optional Namespace field
+	if ar.Request.Namespace != "" {
```
pperiyasamy, line 159: any idea how AdmissionRequest namespace field is derived ?

zshi-redhat: I think it is from webhook server where it knows the k8s object and populates the namespace when sending the AdmissionReview, however I didn't find the exact code the write this field.

pkg/webhook/webhook.go
zshi-redhat: I think both nil or empty slice would work the same, looking at golang doc, it says:
```
In practice, nil slices and empty slices can often be treated in the same way:

    they have zero length and capacity,
    they can be used with the same effect in for loops and append functions,
    and they even look the same when printed.
```

In our case, we use len() to query the length of slice and use for loop against it.
I can change it if you prefer to.

pkg/webhook/webhook.go
pperiyasamy: @zshi-redhat ok, sure, I thought it would use namespace configured from [here](https://github.com/k8snetworkplumbingwg/network-resources-injector/blob/master/deployments/webhook.yaml#L6) :)

pkg/webhook/webhook.go
pperiyasamy: ah! ok :), then let's just use nil as it is.

pkg/webhook/webhook.go
zshi-redhat: okay, it will not :-)
I tried not specifying namespace when creating pod and the AdmissionReview.Request.Namespace was assigned with right namespace by webhook server..

Issue #101, opened 2021-05-27
resource not injected in a large K8s cluster


pperiyasamy:
When we bringup a deployment which has 100 pod replicas, then some pods are not created because NRI doesn't inject required SRIOV VF resources into the pod. 
From troubleshooting NRI (of course logging has to be improved by writing pod info as well in the logs), it looks NRI takes around 15 secs (because of GC ?) to respond to admission request for the failed pod whereas it just look  around 5 secs for the working pods. is admission response ignored due to timeout ? it happens though webhook timeout set to 30 sec (default value?). 
After changing `failurePolicy` from `Ignore` to `Fail`, then all pods are injected with resources. Does Fail policy make any difference ? As per the documentation, it says it would cause pod admission to fail. isn't it ?
If `Fail` policy fixes the issue, then we would need to make `failurePolicy` as configurable parameter, currently it's [hardcoded](https://github.com/k8snetworkplumbingwg/network-resources-injector/blob/628f2982daf036a97ca759efd2a633bd287298ff/pkg/installer/installer.go#L134) to use `Ignore` policy.

pperiyasamy:
The fix #103 with Fail policy seem to working fine, I think #102 and #103 are good to go in.
but still we need to figure out if retries happen upon web hook's timeout (not sure if it's done by helm itself), enable profiling for web hook to find out hot spots.

JanScheurich:
@pperiyasamy: 5 seconds for successfully mutated pods seems already long. But 15 seconds is ages. What is going on there? Where does the NRI spend this time?

pperiyasamy:
@JanScheurich Yes, your suspicion is correct. The retrieval of net-attach-def from k8 for every admission review request causes delay, could reproduce it with single deployment having 100 pod replicas. at times net-attach-def retrieval takes around 5-6 seconds. looks api server imposes rate limiter when requests goes beyond some rate, need to check on that.

can we make NRI to rely on local cache for net-attach-def ? The NRI can watch on net-attach-def events and update its local map, I think `map[string]map[string]string` would be sufficient, containing `namespace + "/" + network-name `as key and `Annotations map `as value.

@martinkennelly @zshi-redhat WDYT ?

zshi-redhat:
[...]

PR #102, opened 2021-06-01
add pod name in the logs

pperiyasamy:
log pod name in info/error logs for better debugging NRI. Issue #101 

Signed-off-by: Periyasamy Palanisamy <periyasamy.palanisamy@est.tech>

PR #103, opened 2021-06-01
configurable failure policy

pperiyasamy:
provide a flexibility to configure failure policy for the web hook, potential fix for the issue #101

### DIFF
--- a/cmd/installer/main.go
+++ b/cmd/installer/main.go
@@ -24,7 +24,7 @@ import (
 func main() {
 	namespace := flag.String("namespace", "kube-system", "Namespace in which all Kubernetes resources will be created.")
 	prefix := flag.String("name", "network-resources-injector", "Prefix added to the names of all created resources.")
-	failurePolicy := flag.String("failure-policy", "Ignore", "failure policy to handle unrecognized errors and timeout errors")
+	failurePolicy := flag.String("failure-policy", "Ignore", "K8 admission controller failure policy to handle unrecognized errors and timeout errors")
 	flag.Parse()
 
 	glog.Info("starting webhook installation")
--- a/cmd/installer/main.go
+++ b/cmd/installer/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"flag"
+
 	"github.com/golang/glog"
 	"github.com/k8snetworkplumbingwg/network-resources-injector/pkg/installer"
 )
@@ -23,8 +24,9 @@ import (
 func main() {
 	namespace := flag.String("namespace", "kube-system", "Namespace in which all Kubernetes resources will be created.")
 	prefix := flag.String("name", "network-resources-injector", "Prefix added to the names of all created resources.")
+	failurePolicy := flag.String("failure-policy", "Ignore", "failure policy to handle unrecognized errors and timeout errors")
 	flag.Parse()
 
 	glog.Info("starting webhook installation")
-	installer.Install(*namespace, *prefix)
+	installer.Install(*namespace, *prefix, *failurePolicy)
 }
--- a/pkg/installer/installer.go
+++ b/pkg/installer/installer.go
@@ -127,11 +127,18 @@ func writeToFile(certificate, key []byte, certFilename, keyFilename string) erro
 	return nil
 }
 
-func createMutatingWebhookConfiguration(certificate []byte) error {
+func createMutatingWebhookConfiguration(certificate []byte, failurePolicyStr string) error {
 	configName := strings.Join([]string{prefix, "mutating-config"}, "-")
 	serviceName := strings.Join([]string{prefix, "service"}, "-")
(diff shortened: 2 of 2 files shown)


pkg/installer/installer.go
```diff
@@ -127,11 +127,18 @@ func writeToFile(certificate, key []byte, certFilename, keyFilename string) erro
...
 }
 
-func createMutatingWebhookConfiguration(certificate []byte) error {
+func createMutatingWebhookConfiguration(certificate []byte, failurePolicyStr string) error {
 	configName := strings.Join([]string{prefix, "mutating-config"}, "-")
 	serviceName := strings.Join([]string{prefix, "service"}, "-")
 	removeMutatingWebhookIfExists(configName)
-	failurePolicy := arv1beta1.Ignore
+	var failurePolicy arv1beta1.FailurePolicyType
+	if failurePolicyStr == "Ignore" {
```
martinkennelly, line 135: can you do case insensitive comparison?

pkg/installer/installer.go
```diff
@@ -127,11 +127,18 @@ func writeToFile(certificate, key []byte, certFilename, keyFilename string) erro
...
+		failurePolicy = arv1beta1.Ignore
+	} else if failurePolicyStr == "Fail" {
```
martinkennelly, line 137: can you do case insensitive comparison?

cmd/installer/main.go
```diff
@@ -16,15 +16,17 @@ package main
...
 	"flag"
+
 	"github.com/golang/glog"
 	"github.com/k8snetworkplumbingwg/network-resources-injector/pkg/installer"
 )
 
 func main() {
 	namespace := flag.String("namespace", "kube-system", "Namespace in which all Kubernetes resources will be created.")
 	prefix := flag.String("name", "network-resources-injector", "Prefix added to the names of all created resources.")
+	failurePolicy := flag.String("failure-policy", "Ignore", "failure policy to handle unrecognized errors and timeout errors")
```
martinkennelly, line 27: Can you extend the description to say 

> K8 admission controller failure policy to handle unrecognized errors and timeout errors

martinkennelly:
@pperiyasamy I see this patch is depending/containing a commit from PR #102. 
@zshi-redhat I think failpolicy should be fail by default. I'd like to hear your opinion on this.
@pperiyasamy If so, we may need to update our deployment k8 [artifact ](https://github.com/k8snetworkplumbingwg/network-resources-injector/blob/master/deployments/server.yaml)as well but please wait until Zenghui has weighted in.

zshi-redhat:
> @pperiyasamy I see this patch is depending/containing a commit from PR #102.
> @zshi-redhat I think failpolicy should be fail by default. I'd like to hear your opinion on this.
> @pperiyasamy If so, we may need to update our deployment k8 [artifact ](https://github.com/k8snetworkplumbingwg/network-resources-injector/blob/master/deployments/server.yaml)as well but please wait until Zenghui has weighted in.

@martinkennelly From [here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy), it says admission V1 uses `Fail` as default `FailurePolicy`, admission V1beta uses `Ignore` as default `FailurePolicy`. 

Currently we are still using V1beta in NRI, maybe `Ignore` could be the default? We can switch the default to `Fail` once moving to admission V1. wdyt?

martinkennelly:
> > @pperiyasamy I see this patch is depending/containing a commit from PR #102.
> > @zshi-redhat I think failpolicy should be fail by default. I'd like to hear your opinion on this.
> > @pperiyasamy If so, we may need to update our deployment k8 [artifact ](https://github.com/k8snetworkplumbingwg/network-resources-injector/blob/master/deployments/server.yaml)as well but please wait until Zenghui has weighted in.
> 
> @martinkennelly From [here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy), it says admission V1 uses `Fail` as default `FailurePolicy`, admission V1beta uses `Ignore` as default `FailurePolicy`.
> 
> Currently we are still using V1beta in NRI, maybe `Ignore` could be the default? We can switch the default to `Fail` once moving to admission V1. wdyt?

Missed that. Thanks!
It seems reasonable to upgrade to V1 with #96 and get fail by default but I like this patch to make failure configurable. This is new behavior for NRI so we need to make it clear in release notes.

zshi-redhat (approved):

cmd/installer/main.go
pperiyasamy: done.

pkg/installer/installer.go
pperiyasamy: done.

pkg/installer/installer.go
pperiyasamy: done.

martinkennelly (approved):

PR #98, opened 2021-05-04
E2E tests for numerous NRI features

martinkennelly:
E2E Tests for features:
  * node selector
  * hugepages (optional)
  * user defined injections
  * resourceName

Co-authored-by: Michal Guzieniuk <michalx.guzieniuk@intel.com>
Co-authored-by: Martin Kennelly <martin.kennelly@intel.com>

Hugepages test is optional and will be skipped if the number of 1Gi / 2Mi hugepages arent available. This allows us to run the same tests in an environment with (self hosted) and without hugepages (GH VM).

This patch adds a self hosted job which requires a GH environment. A GH environment can contain a set of approvers to allow execution. This environment should be created before this patch can be merged and tested.

Example of how we can use GH environments to allow a set of people to approve workflows: [https://docs.google.com/document/d/1kWko-Hnne-BcuFvzEZPlmGxFQ9FSkTzPSzxP1m9xUuY/edit](https://docs.google.com/document/d/1kWko-Hnne-BcuFvzEZPlmGxFQ9FSkTzPSzxP1m9xUuY/edit)

Thanks to @MichalGuzieniuk for creating all the tests.

Sample execution: https://github.com/martinkennelly/network-resources-injector/runs/2504160518?check_suite_focus=true

PR #104, opened 2021-06-01
Allow multiple add patch Ops for UDI annotations

### LINKED ISSUE #94: User Defined Injections - injects only one property
For below ConfigMap, where there is more than one key/value pair NRI injects only one pair into POD specification. I observed that this injection is random. It means that only one of the pair: "top-secret" or "k8s.v1.cni.cncf.io/network" is added to the POD.

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: nri-user-defined-injections
  namespace: kube-system
data:
  "customInjection": '{"op": "add", "path": "/metadata/annotations", "value": {"k8s.v1.cni.cncf.io/networks": "sriov-net-attach-def"}}'
  "secondInjection": '{"op": "add", "path": "/metadata/annotations", "value": {"top-secret": "password"}}'
```

POD specification
```
apiVersion: v1
kind: Pod
metadata:
  name: testpod
  labels:
    customInjection: "true"
    secondInjection: "true"
  annotations:
    k8s.v1.cni.cncf.io/networks: foo-network
spec:
  containers:
  - name: app
    image: alpine
    command: [ "/bin/sh", "-c", "sleep INF" ]

```

Logs:
```
I0413 10:45:43.523737       1 main.go:69] starting mutating admission controller for network resources injection
I0413 10:45:43.524389       1 tlsutils.go:120] added client CA to cert pool from path '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt'
I0413 10:45:43.524405       1 tlsutils.go:122] added '1' client CA(s) to cert pool
I0413 10:46:13.550537       1 webhook.go:943] Initializing user-defined injections with key: customInjection, value: {"op": "add", "path": "/metadata/annotations", "value": {"k8s.v1.cni.cncf.io/networks": "sriov-net-attach-def"}}
I0413 10:46:13.550584       1 webhook.go:943] Initializing user-defined injections with key: secondInjection, value: {"op": "add", "path": "/metadata/annotations", "value": {"top-secret": "password"}}
I0413 10:46:58.393901       1 webhook.go:704] Received mutation request
I0413 10:46:58.438718       1 webhook.go:677] search v1.multus-cni.io/default-network in original pod annotations
I0413 10:46:58.438742       1 webhook.go:684] search v1.multus-cni.io/default-network in user-defined injections
I0413 10:46:58.438751       1 webhook.go:698] v1.multus-cni.io/default-network is not found in either pod annotations or user-defined injections
I0413 10:46:58.438761       1 webhook.go:677] search k8s.v1.cni.cncf.io/networks in original pod annotations
I0413 10:46:58.438769       1 webhook.go:680] k8s.v1.cni.cncf.io/networks is defined in original pod annotations
I0413 10:46:58.438795       1 webhook.go:256] 'foo-network' is not in JSON format: invalid character 'o' in literal false (expecting 'a')... trying to parse as comma separated network selections list
I0413 10:46:58.450875       1 webhook.go:354] network attachment definition 'default/foo-network' found
I0413 10:46:58.450915       1 webhook.go:361] resource 'example.com/foo' needs to be requested for network 'default/foo-network'
I0413 10:46:58.450928       1 webhook.go:790] honor-resources=false
I0413 10:46:58.450955       1 webhook.go:800] injectHugepageDownApi=true
I0413 10:46:58.450981       1 webhook.go:858] patch after all mutations: [{add /spec/containers/0/resources/requests map[]} {add /spec/containers/0/resources/limits map[]} {add /spec/containers/0/resources/requests/example.com~1foo {{1 0} {<nil>}  DecimalSI}} {add /spec/containers/0/resources/limits/example.com~1foo {{1 0} {<nil>}  DecimalSI}} {add /spec/containers/0/volumeMounts/- {podnetinfo false /etc/podnetinfo  <nil> }} {add /spec/volumes/- {podnetinfo {nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil &DownwardAPIVolumeSource{Items:[]DownwardAPIVolumeFile{DownwardAPIVolumeFile{Path:labels,FieldRef:&ObjectFieldSelector{APIVersion:,FieldPath:metadata.labels,},ResourceFieldRef:nil,Mode:nil,},DownwardAPIVolumeFile{Path:annotations,FieldRef:&ObjectFieldSelector{APIVersion:,FieldPath:metadata.annotations,},ResourceFieldRef:nil,Mode:nil,},},DefaultMode:nil,} nil nil nil nil nil nil nil nil nil nil nil nil}}} {add /metadata/annotations map[k8s.v1.cni.cncf.io/networks:sriov-net-attach-def]} {add /metadata/annotations map[k8s.v1.cni.cncf.io/networks:foo-network top-secret:password]}]
I0413 10:46:58.451457       1 webhook.go:396] sending response to the Kubernetes API server
I0413 10:47:13.565255       1 webhook.go:952] Removing stale entry: customInjection from user-defined injections
I0413 10:47:13.565289       1 webhook.go:952] Removing stale entry: secondInjection from user-defined injections
```

martinkennelly:
Grouping multiple 'add' op patch's for annotations allows us to deterministically override pod annotations.
Get the union of add op patches and then only apply existing pod annotations if there is no conflict.

Also, restrict to only allow "add" Op for annotations

Fixes #94 
Tested with #98

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -639,22 +639,6 @@ func getResourceList(resourceRequests map[string]int64) *corev1.ResourceList {
 	return &resourceList
 }
 
-func appendPodAnnotation(patch []jsonPatchOperation, pod corev1.Pod, userDefinedPatch jsonPatchOperation) []jsonPatchOperation {
-	annotMap := make(map[string]string)
-	for k, v := range pod.ObjectMeta.Annotations {
-		annotMap[k] = v
-	}
-	for k, v := range userDefinedPatch.Value.(map[string]interface{}) {
-		annotMap[k] = v.(string)
-	}
-	patch = append(patch, jsonPatchOperation{
-		Operation: "add",
-		Path:      "/metadata/annotations",
-		Value:     annotMap,
-	})
-	return patch
-}
-
 func createCustomizedPatch(pod corev1.Pod) ([]jsonPatchOperation, error) {
 	var userDefinedPatch []jsonPatchOperation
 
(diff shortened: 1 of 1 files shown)


zshi-redhat:
@martinkennelly thanks for the fix!
IIUC, the issue was if there are multiple patches to `/metadata/annotations` in the mutating webhook response, the last patch would override previous patches (even the operation is `add`), right?

martinkennelly:
@zshi-redhat 
>the last patch would override previous patches (even the operation is add), right?

Exactly, yes. So, we either limit the number of 'add' injections for annotations to exactly one, or we group the 'add' ops for annotations like in this patch and emit logs if there are conflicts when attempting to group them. 
Ill let you decide on whether we limit or group. 

Its much simpler to limit the number of 'add' ops for annotations to one and maybe that's sufficient for UDI. I think this patch gives more flexibility for UDI config designers however.

zshi-redhat:
> @zshi-redhat
> 
> > the last patch would override previous patches (even the operation is add), right?
> 
> Exactly, yes. So, we either limit the number of 'add' injections for annotations to exactly one, or we group the 'add' ops for annotations like in this patch and emit logs if there are conflicts when attempting to group them.
> Ill let you decide on whether we limit or group.
> 
I think we should allow multiple `add` injections, this patch is well done!

zshi-redhat (approved):

PR #106, opened 2021-06-10
make webhook to use local nad cache

### LINKED ISSUE #101: resource not injected in a large K8s cluster

pperiyasamy:
this implements local cache using network-attachment-definition-client's informer, web hook looks up this cache, fall back to K8 api when entry is not found.

Fixes #101 

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -15,23 +15,24 @@
 package main
 
 import (
-	"crypto/tls"
 	"context"
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"net/http"
 	"os"
 	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"github.com/fsnotify/fsnotify"
 	"github.com/golang/glog"
+	netcache "github.com/k8snetworkplumbingwg/network-resources-injector/pkg/tools"
 	"github.com/k8snetworkplumbingwg/network-resources-injector/pkg/webhook"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	defaultClientCa              = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	defaultClientCa               = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 	userDefinedInjectionConfigMap = "nri-user-defined-injections"
 )
 
@@ -89,6 +90,11 @@ func main() {
 	if err != nil {
 		glog.Fatalf("error in setting resource name keys: %s", err.Error())
 	}
+
+	//initiaze webhook with cache
+	netAnnotationCache := netcache.Create()
+	netAnnotationCache.Start()
+	webhook.SetNetAttachDefCache(netAnnotationCache)
 
 	go func() {
 		/* register handlers */
@@ -193,4 +199,7 @@ func main() {
 			webhook.SetCustomizedInjections(cm)
 		}
 	}
+
+	// TODO: find a way to stop cache, should we run the above block in a go routine and make main module
+	// to respond to terminate singal ?
 }
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -91,7 +91,7 @@ func main() {
 		glog.Fatalf("error in setting resource name keys: %s", err.Error())
 	}
 
-	//initiaze webhook with cache
+	//initialize webhook with cache
 	netAnnotationCache := netcache.Create()
 	netAnnotationCache.Start()
 	webhook.SetNetAttachDefCache(netAnnotationCache)
--- a/pkg/tools/cache.go
+++ b/pkg/tools/cache.go
@@ -16,12 +16,13 @@ package cache
 
 import (
 	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/golang/glog"
 	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned"
 	"github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/informers/externalversions"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 )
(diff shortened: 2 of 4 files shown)


cmd/webhook/main.go
```diff
@@ -193,4 +199,7 @@ func main() {
 			webhook.SetCustomizedInjections(cm)
 		}
 	}
+
+	// TODO: find a way to stop cache, should we run the above block in a go routine and make main module
```
pperiyasamy, line 203: @martinkennelly @zshi-redhat shall I make to run above for block into a goroutine, exit gracefully on term signal ? this change would also enable to stop nad cache informer.

or is the above for block intentional ?

pperiyasamy: looks this issue is being addressed in this [PR](https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/62), will update it here once that PR is merged.

zshi-redhat (approved):

pkg/tools/cache.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (nc *NetAttachDefCache) put(networkName string, annotations map[string]string) {
+	nc.networkAnnotationsMapMutex.Lock()
+	nc.networkAnnotationsMap[networkName] = annotations
+	nc.networkAnnotationsMapMutex.Unlock()
+}
+
+// Get returns annotations map for the given network name, if it's not available
+// return nil
+func (nc *NetAttachDefCache) Get(networkName string) map[string]string {
```
martinkennelly, line 94: small nit - feel free to ignore: up to you if you want to change it - it would be better to have two arguments here  I think - namespace and nad name, then you can bury the joining of the two strings ```net.Namespace + "/" + net.Name``` in this function instead of having to do it consistently when calling it like  I see below.  WDYT?

pkg/webhook/webhook.go
```diff
@@ -367,18 +369,24 @@ func getNetworkAttachmentDefinition(namespace, name string) (*cniv1.NetworkAttac
...
 func parseNetworkAttachDefinition(net *multus.NetworkSelectionElement, reqs map[string]int64, nsMap map[string]string) (map[string]int64, map[string]string, error) {
 	/* for each network in annotation ask API server for network-attachment-definition */
-	networkAttachmentDefinition, err := getNetworkAttachmentDefinition(net.Namespace, net.Name)
-	if err != nil {
-		/* if doesn't exist: deny pod */
-		reason := errors.Wrapf(err, "could not find network attachment definition '%s/%s'", net.Namespace, net.Name)
-		glog.Error(reason)
-		return reqs, nsMap, reason
+	var annotationsMap map[string]string
+	annotationsMap = nadCache.Get(net.Namespace + "/" + net.Name)
```
martinkennelly, line 373: nit: why not ":="  save one line :P

pkg/tools/cache.go
```diff
@@ -0,0 +1,120 @@
...
+	Stop()
+	Get(string) map[string]string
+}
+
+func Create() NetAttachDefCacheService {
+	return &NetAttachDefCache{make(map[string]map[string]string),
+		&sync.Mutex{}, make(chan struct{})}
+}
+
+// Start start the informer for NetworkAttachmentDefinition, upon events populate the cache
```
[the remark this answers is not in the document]
pperiyasamy: done

pkg/tools/cache.go
```diff
@@ -0,0 +1,120 @@
...
+func (nc *NetAttachDefCache) Start() {
+	factory := externalversions.NewFilteredSharedInformerFactory(setupNetAttachDefClient(), 0, "", func(o *metav1.ListOptions) {})
```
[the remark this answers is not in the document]
pperiyasamy: done

pkg/tools/cache.go
```diff
@@ -0,0 +1,120 @@
...
+			netAttachDef := obj.(*cniv1.NetworkAttachmentDefinition)
+			nc.remove(netAttachDef.Namespace + "/" + netAttachDef.Name)
+		},
+	})
+	go informer.Run(nc.stopper)
+}
+
+// Stop stop the NetworkAttachmentDefinition informer
+func (nc *NetAttachDefCache) Stop() {
+	close(nc.stopper)
```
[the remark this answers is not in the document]
pperiyasamy: good find! now added a code for graceful shutdown of nad cache with atomic isRunning variable, hope that's ok now.

pkg/tools/cache.go
pperiyasamy: yes, it makes sense. done.

pkg/tools/cache.go
```diff
@@ -0,0 +1,120 @@
...
+	defer nc.networkAnnotationsMapMutex.Unlock()
+	if annotationsMap, exists := nc.networkAnnotationsMap[networkName]; exists {
+		return annotationsMap
+	}
+	return nil
+}
+
+func (nc *NetAttachDefCache) remove(networkName string) {
+	nc.networkAnnotationsMapMutex.Lock()
+	delete(nc.networkAnnotationsMap[networkName], networkName)
```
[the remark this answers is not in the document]
pperiyasamy: done.

pkg/webhook/webhook.go
pperiyasamy: done

pkg/tools/cache.go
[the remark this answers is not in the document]
martinkennelly: @pperiyasamy I don't understand how atomic.StoreInt32 solves the issue. I don't think the issue with the informer is possible to solve without the informer signally back that it has completed/ended and there isn't any way to do this, it seems.

pkg/tools/cache.go
```diff
@@ -0,0 +1,147 @@
...
+			defer mutex.Unlock()
+			netAttachDef := obj.(*cniv1.NetworkAttachmentDefinition)
+			nc.remove(netAttachDef.Namespace, netAttachDef.Name)
+		},
+	})
+	go func() {
+		atomic.StoreInt32(&(nc.isRunning), int32(1))
+		// informer Run blocks until informer is stopped
+		glog.Infof("starting net-attach-def informer")
+		informer.Run(nc.stopper)
```
martinkennelly, line 85: Can we put in a log after .Run "net-attach-def informer stopped"? It may aid debugging.

martinkennelly (commented):
LGTM - just small change required. Thanks Peri for the changes!!

pkg/tools/cache.go
pperiyasamy: done

pkg/tools/cache.go
```diff
@@ -0,0 +1,147 @@
...
+		atomic.StoreInt32(&(nc.isRunning), int32(0))
+	}()
+}
+
+// Stop teardown the NetworkAttachmentDefinition informer
+func (nc *NetAttachDefCache) Stop() {
+	close(nc.stopper)
+	func(limit time.Duration) {
```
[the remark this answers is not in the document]
pperiyasamy: yes :) done.

martinkennelly (approved):
I have verified this change with #98 ! Thanks @pperiyasamy for all your hard work and changes here! I am waiting on @zshi-redhat to review this now as there are changes since he last approved.

pkg/tools/cache.go
```diff
@@ -0,0 +1,145 @@
...
+		informer.Run(nc.stopper)
+		glog.Infof("net-attach-def informer is stopped")
+		atomic.StoreInt32(&(nc.isRunning), int32(0))
+	}()
+}
+
+// Stop teardown the NetworkAttachmentDefinition informer
+func (nc *NetAttachDefCache) Stop() {
+	close(nc.stopper)
+	tEnd := time.Now().Add(3 * time.Second)
```
MichalGuzieniuk, line 94: Could you explain why you choose 3 seconds? Is this enough to stop net-attach-def informer?

pperiyasamy: @MichalGuzieniuk thanks for looking at this. I just chose 3 seconds based on Kubelet's default shutdown grace period (i.e. 30 sec) for a pod. but, yes, if net-attach-def events are not yet flushed away from informer then it might take a while for informer to return after close. 
there is no much activity happening in this cache module apart from updating its local map, do you see this as a problem or any other concern ?

MichalGuzieniuk: Fine for me. I was just interested if you choose this value because in tests you discovered that is the best or had other reason behind.

zshi-redhat:
> I have verified this change with #98 ! Thanks @pperiyasamy for all your hard work and changes here! I am waiting on @zshi-redhat to review this now as there are changes since he last approved.

Thanks for all the changes and reviews!

Issue #61, opened 2021-01-18
How can we better handle feature control switches in NRI?


martinkennelly:
Currently in pkg/webhook we have defined variables to control whether certain functionality is enabled/disabled:

```
var (
	clientset              kubernetes.Interface
	injectHugepageDownApi  bool
	resourceNameKeys       []string
	honorExistingResources bool
)
```

As NRI feature-set grows, this approach is unworkable. How can we prevent the growth of variable feature control switches in NRI?

I think it would be better to define a struct or interface and pass it to functions.

pperiyasamy:
Looks config map would be a good fit for passing these config options like proposal done for issue #57, we could reuse the same config map with adding config.json as below:

```
apiVersion: v1
data:
  network-resource-injector-environment-variable: '{"spec":{"containers":[{"image":"docker.io/nfvpe/sriov-device-plugin","env":[{"name":"APP_ROLE","value":"client"}],"name":"application"}]}}'
  config.json: |
    {
	  "webhookconfig": {
	    "injectHugepageDownApi": true,
		"networkResourceNameKeys": ["k8s.v1.cni.cncf.io/resourceName", "k8s.v1.cni.cncf.io/bridgeName"],
		"honorResources": false
	  }
	}
kind: ConfigMap
metadata:
  name: network-resource-injector-customized-injection-config
  namespace: network-resource-injector-namespace
```

and mount config.json as config-volume inside the NRI pod, load it into a struct and configure the webhook accordingly.

zshi-redhat:
I think the issue has two aspects, one is to pass config to NRI, the other is to have structured config for various features.

For passing config to NRI,  as @pperiyasamy mentioned, we could use configMap, or environment variable or a regular config file (mounted via volume) or cli cmd options. configMap and regular config file have advantage that features can be toggled on the fly when NRI is running,  while environment variable and cli cmd can only be configured when NRI starts. If all of them are to be supported, there needs to be an order of preference.

For stuctured config data, as feature grows, we probably can define sections to organize the configurations, just like ini file, which may contain multiple sections, with each section contains key:value pair. For example:

```
[default]
enableCustomizedInjection = false

[resource]
injectResourceName = true
honorExistingResources = true
resourceNameKeys = ["key 1", "key 2"]

[volume] or [downward]
injectHugepageDownApi = true
injectNetInfo = true

```

Section name can be different, depending on whether we want to separate them or have single section.

[...]

PR #108, opened 2021-06-17
Add controlSwitches package to be able to toggle feature state at runtime

MichalGuzieniuk:
The aim of this pull request is to solve problem describe within issue https://github.com/k8snetworkplumbingwg/network-resources-injector/issues/61 

This pull request brings first part of new configMap support - so the handing of feature state in runtime. With new configMap described wider in **readme** it is possible to change initial feature state without need to NRI recreation.

The implementation is done within controlswitches package. In result, some parts of code were moved from webhook.go file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
       * [Disable adding client CAs to server TLS endpoint](#disable-adding-client-cas-to-server-tls-endpoint)
       * [Client CAs](#client-cas)
    * [Additional features](#additional-features)
+      * [Features control switches](#features-control-switches)
       * [Expose Hugepages via Downward API](#expose-hugepages-via-downward-api)
       * [Node Selector](#node-selector)
       * [User Defined Injections](#user-defined-injections)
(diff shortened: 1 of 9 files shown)


zshi-redhat:
@MichalGuzieniuk thanks for the patch! It looks generally good to me.

One question: do you have plan to remove userDefinedInjection configMap and add it in the `nri-control-switches` configMap?
I see that resourceNameKeys doesn't have a enablement flag, the code checks the len of the map to determine whether resourceNameKeys is configured or not, perhaps we can do the same for userDefinedInjection, wdyt?

MichalGuzieniuk:
@zshi-redhat I is a good idea to move userDefinedInjection configMap to `nri-control-switches` configMap. Such structure was agreed within https://github.com/k8snetworkplumbingwg/network-resources-injector/issues/61 But I prefer to make small changes instead of one big patch. So, as a next step I can move user injections.

I talked with @martinkennelly and we decide that for resourceNameKeys flag is not required because NRI without resourceNameKeys will not work.

On the other hand, we have userDefinedInjection, feature that is enabled by default and controlled with second configMap. You are right that I could use length of map to determine feature state. Right now additional flag doubles configuration. I will remove that flag in next commit.

zshi-redhat:
> @zshi-redhat I is a good idea to move userDefinedInjection configMap to `nri-control-switches` configMap. Such structure was agreed within #61 But I prefer to make small changes instead of one big patch. So, as a next step I can move user injections.

ack+
> 
> I talked with @martinkennelly and we decide that for resourceNameKeys flag is not required because NRI without resourceNameKeys will not work.

Yes, resourceNameKeys must have a default value.
> 
> On the other hand, we have userDefinedInjection, feature that is enabled by default and controlled with second configMap. You are right that I could use length of map to determine feature state. Right now additional flag doubles configuration. I will remove that flag in next commit.

sounds good.

zshi-redhat:
@MichalGuzieniuk this PR needs rebase.

zshi-redhat:
@MichalGuzieniuk could you rebase again, seems we have an additional commit that doesn't belong to this PR: `Merge branch 'master' into issue_61`.

MichalGuzieniuk:
> @MichalGuzieniuk could you rebase again, seems we have an additional commit that doesn't belong to this PR: `Merge branch 'master' into issue_61`.

@zshi-redhat My bad, I did merge instead of rebase. Fixed.

zshi-redhat (approved):

MichalGuzieniuk:
@zshi-redhat Could you merge this PR? Unfortunately from our side only @martinkennelly was able to do it but he is on holidays.

Issue #111, opened 2021-08-10
Certificates v1beta1 will no longer be support on k8s release 1.22


Eoghan1232:
K8s release 1.22 onwards, it looks like k8s.io/api/certificates/v1beta1 will not be supported.
NRI will have issues once this happens. 

Should we begin upgrading from v1beta1 to v1 for the certificates?

Recently, admission was switched from v1beta1 to v1. 
https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/96/files

VivekThrivikraman-est:
From API version certificates.k8s.io/v1 onwards signerName is a required key for CSR's(certificate signing requests) and also they have made multiple built in signers for specific purposes which are supposed to be used for only k8s internal services.
Below are the few options which can be considered :

1. Use k8s internal kubelet-serving signer to sign nri webhook certificate, by making changes to the certificate request as required by that signer(like common name start with system:node etc).
2. Use custom signers like cert-manager(https://cert-manager.io/).
3. Use self signed certificate.

But all the above options have some disadvantages Option1: we would be using signer of kubelet  Option2:Cert manager's support for k8s csr API is currently experimental and also this would add an additional dependency Option3:Is not normally used in production(but since it's all in cluster should this be ok?).

adrianchiris:
some more info on the deprecation:

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#certificatesigningrequest-v122

https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes

adrianchiris:
Summary from today's discussion about this issue

Today, NRI supports either providing a signed certificate from an external source using `tls-cert-file`, `tls-private-key-file`
or by generating a certificate request and signing it using k8s signer via k8s certificates/v1beta1 API CertificateSigningRequest object (self approved by NRI).

The latter is no longer available in Kubernetes 1.22. 
in k8s certificates/v1 a signer need to be provided.

see: https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes

Kubernetes provides [built-in signers](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#signers) however it seems that none of them fit the bill.

[...]

PR #114, opened 2021-09-09
Remove usage of k8s api certificates/v1beta1 and use self signed certificate

VivekThrivikraman-est:
K8s release 1.22 onwards k8s.io/api/certificates/v1beta1 will not be supported and with certificates.k8s.io/v1 the k8s signers are designed to be used only for k8s internal services. This PR introduces self signed certificates for NRI webhook server instead of using k8s signers.

Fixes: #111


adrianchiris (approved):
Thank you for your contribution !

Code Looks OK IMO, I did not test this though.

e0ne (approved):

Issue #119, opened 2021-10-19
NRI pod restart not handled 


VivekThrivikraman-est:
NRI doesn't currently handle nri pod restarts, if for some reason nri pod gets killed and there are no other nri pods in the cluster, then nri pod(and other pods) would not be able to start as the service serving the nri webhook(which is the nri pod itself) is down(as webhook configuration is still present).

zshi-redhat:
@VivekThrivikraman-est  Will it work for you if it uses deployment rather than pod to create NRI instances?

PR #120, opened 2021-10-21
Handle nri pod restarts

VivekThrivikraman-est:
This PR avoids nri webhook service from being invoked for nri pods and any other pod in kube-system. 

Fixes: #119

### DIFF
--- a/deployments/webhook.yaml
+++ b/deployments/webhook.yaml
@@ -14,6 +14,12 @@ webhooks:
         namespace: ${NAMESPACE}
         path: "/mutate"
       caBundle: ${CA_BUNDLE}
+    namespaceSelector:
+      matchExpressions:
+        - key: "kubernetes.io/metadata.name"
+          operator: "NotIn"
+          values:
+            - "kube-system"
     rules:
       - operations: [ "CREATE" ]
         apiGroups: ["apps", ""]
--- a/pkg/installer/installer.go
+++ b/pkg/installer/installer.go
@@ -107,6 +107,19 @@ func createMutatingWebhookConfiguration(certificate []byte, failurePolicyStr str
 	}
 	sideEffects := arv1.SideEffectClassNone
 	path := "/mutate"
+	namespaces := []string{"kube-system"}
+	if namespace != "kube-system" {
+		namespaces = append(namespaces, namespace)
+	}
+	namespaceSelector := metav1.LabelSelector{
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      "kubernetes.io/metadata.name",
+				Operator: metav1.LabelSelectorOpNotIn,
+				Values:   namespaces,
+			},
+		},
+	}
 	configuration := &arv1.MutatingWebhookConfiguration{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: configName,
@@ -128,6 +141,7 @@ func createMutatingWebhookConfiguration(certificate []byte, failurePolicyStr str
 				FailurePolicy:           &failurePolicy,
 				AdmissionReviewVersions: []string{"v1"},
 				SideEffects:             &sideEffects,
+				NamespaceSelector:       &namespaceSelector,
 				Rules: []arv1.RuleWithOperations{
 					arv1.RuleWithOperations{
 						Operations: []arv1.OperationType{arv1.Create},
--- a/README.md
+++ b/README.md
@@ -158,6 +158,8 @@ Currently supported arguments are below. If needed, detailed description is avai
 |network-resource-name-keys|k8s.v1.cni.cncf.io/resourceName|comma separated resource name keys|YES|
 |honor-resources|false|Honor the existing requested resources requests & limits|YES|
 
+NOTE: Network Resource Injector would not mutate pods in kube-system namespace.
+
 ### Features control switches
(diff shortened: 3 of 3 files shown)


VivekThrivikraman-est:
> @VivekThrivikraman-est Thanks for the fix! Would you mind adding a piece of doc in the README.md file? Just doc the behavior that NRI would not mutate pods in the kube-system namespace, maybe in the additional features section. wdyt?

In the last commit I have updated the README.md to reflect the change.

> @VivekThrivikraman-est Thanks for the fix! Would you mind adding a piece of doc in the README.md file? Just doc the behavior that NRI would not mutate pods in the kube-system namespace, maybe in the additional features section. wdyt?

In the last commit I have updated the README.md.

pkg/installer/installer.go
```diff
@@ -107,6 +107,15 @@ func createMutatingWebhookConfiguration(certificate []byte, failurePolicyStr str
 	}
 	sideEffects := arv1.SideEffectClassNone
 	path := "/mutate"
+	namespaceSelector := metav1.LabelSelector{
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      "kubernetes.io/metadata.name",
+				Operator: metav1.LabelSelectorOpNotIn,
+				Values:   []string{"kube-system"},
```
adrianchiris, line 115: OK, so here is what i found.

1. it is recommended to avoid operating on kube-system namespace [1]
2. it is recommended to avoid operating in the namespace the webhook is deployed to avoid deadlocks [2]

these two namespaces might not be the same (by default here they are). you should also add the namespace passed to `installer.Install()` if its different than "kube-system"

[1] https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#avoiding-operating-on-the-kube-system-namespace
[2] https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#avoiding-deadlocks-in-self-hosted-webhooks

deployments/webhook.yaml
```diff
@@ -14,6 +14,12 @@ webhooks:
         namespace: ${NAMESPACE}
         path: "/mutate"
       caBundle: ${CA_BUNDLE}
+    namespaceSelector:
```
adrianchiris, line 17: unrelated comment to this change:

@zshi-redhat i see that in sriov-network-operator we are deploying this with `failurePolicy: Ignore`
which is why i think we never faced the issue when deploying with the operator.

should we align here (separate PR ofc) ?

adrianchiris (changes requested):
Thanks for working on this @VivekThrivikraman-est  ! 

added some comments.

pkg/installer/installer.go
adrianchiris: @VivekThrivikraman-est apart of this i think we are good.

VivekThrivikraman-est: @adrianchiris "namespace" passed to installer seems to be used for creation of service . The namespace of webhook seems to be directly specified in server.yaml (as kube-system). Should the installer container(init container) get the namespace internally(through env variables like nri container currently does) and add that to the list ? I think that would be required if user changes the namespace in server.yaml and also the auth.yaml (to have nri pod in another namespace).

adrianchiris: > Should the installer container(init container) get the namespace internally(through env variables like nri container currently does) and add that to the list ?

it might, but thats not related to this change IMO. you could also pass `--namespace=${NAMESPACE}` in initcontainer args. `NAMESPACE` being an environment variable containing the pod's namespace .

my ask here is IF the user decided to change where webhook and its components deploy (server.yaml, auth.yaml)
and provided the correct value for the namespace flag  to Installer, then this bit should take it into account.

VivekThrivikraman-est: > my ask here is IF the user decided to change where webhook and its components deploy (server.yaml, auth.yaml)
and provided the correct value for the namespace flag to Installer, then this bit should take it into account.

Ok will make that change and push.

adrianchiris (approved):

deployments/webhook.yaml
zshi-redhat: @adrianchiris In sriov-network-operator, it sets the `failurePolicy` to `Ignore` upon sriov-operator pod exist and reset (automatically by operator reconcilation) `failurePolicy` to `Fail` when sriov-operator pod is back/reinstalled.

zshi-redhat: Code change in sriov-network-operator: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/177

pkg/installer/installer.go
zshi-redhat: > @adrianchiris "namespace" passed to installer seems to be used for creation of service . The namespace of webhook seems to be directly specified in server.yaml (as kube-system). Should the installer container(init container) get the namespace internally(through env variables like nri container currently does) and add that to the list ? 

@VivekThrivikraman-est  Yes, I think it should. (could be in a separate PR)
Just for my understanding: How did you create the webhook.yaml and server.yaml? It looks to me that we would need either webhook.yaml or installer initContainer, but not both since they do the same webhook creation.

VivekThrivikraman-est: > It looks to me that we would need either webhook.yaml or installer initContainer, but not both since they do the same webhook creation.

@zshi-redhat yes it would be only one required, I made changes so that it is consistent both ways.

zshi-redhat (approved):

zshi-redhat:
@VivekThrivikraman-est @adrianchiris I merged this PR per two approvals. 
Feel free to propose other PRs to fix the inconsistent of installer and webhook.yaml or any other potential issues you may see.

PR #121, opened 2021-11-25
E2E tests for numerous NRI features

MichalGuzieniuk:
E2E Tests for features:

- node selector
- hugepages (optional)
- user defined injections
- resourceName

Co-authored-by: Michal Guzieniuk michalx.guzieniuk@intel.com
Co-authored-by: Martin Kennelly martin.kennelly@intel.com

Hugepages test is optional and will be skipped if the number of 1Gi / 2Mi hugepages arent available. This allows us to run the same tests in an environment with (self hosted) and without hugepages (GH VM).

This patch adds a self hosted job which requires a GH environment. A GH environment can contain a set of approvers to allow execution. This environment should be created before this patch can be merged and tested.

Example of how we can use GH environments to allow a set of people to approve workflows: https://docs.google.com/document/d/1kWko-Hnne-BcuFvzEZPlmGxFQ9FSkTzPSzxP1m9xUuY/edit

Thanks to @MichalGuzieniuk for creating all the tests.

Sample execution: https://github.com/martinkennelly/network-resources-injector/runs/2504160518?check_suite_focus=true

**NOTE**: This PR is a continuation of https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/98 but I was not able to directly resolved there conflicts.


test/e2e/custominjection_test.go
```diff
@@ -0,0 +1,531 @@
+package e2e
```
zshi-redhat, line 1: could you change the test file name to userdefined_test.go?

test/e2e/custominjection_test.go
```diff
@@ -0,0 +1,531 @@
...
+			util.DeleteNetworkAttachmentDefinition(networkClient.K8sCniCncfIoV1Interface, "sriov-net-attach-def", nad2, timeout)
+
+			if nil != configMap {
+				util.DeleteConfigMap(cs.CoreV1Interface, configMap, timeout)
+				configMap = nil
+			}
+		})
+
+		It("Config map in correct namespace, one label to inject, POD specification WITHOUT resource name", func() {
+			configMap = util.GetConfigMap("nri-user-defined-injections", "kube-system")
```
zshi-redhat, line 45: I guess we shall wait for the controlSwitch configmap PR to merge first, then update this configMpa name and the following test cases? Or do you prefer the reverse order?

scripts/e2e_setup_cluster.sh
```diff
@@ -142,18 +153,18 @@ retry kubectl create -f "${CNIS_DAEMONSET_URL}"
 retry kubectl -n kube-system wait --for=condition=ready -l name="${CNIS_NAME}" pod --timeout=300s
 echo "## install NRI"
 retry kubectl create -f "${root}/deployments/auth.yaml"
-retry kubectl create -f "${root}/deployments/server.yaml"
+retry kubectl create -f "${root}/deployments/server_huge.yaml"
```
zshi-redhat, line 156: Maybe we can reuse the existing server.yaml and use controlSwitch to toggle the `injectHugepageDownApi` configuration for testing?

.github/workflows/e2e.yml
```diff
@@ -1,8 +1,8 @@
 name: e2e
-on: [push, pull_request]
+on: [pull_request]
 jobs:
-  e2e-test:
-    name: E2E test
+  e2e-test-cloud:
```
zshi-redhat, line 4: Do we want to test with a higher go-version? like 1.15 or 1.16?

test/e2e/custominjection_test.go
MichalGuzieniuk: Yes, I think that we can wait with this change for the second PR merge. For sure e2e code may require some adaptations.

.github/workflows/e2e.yml
MichalGuzieniuk: Yes, I think that we can verify it with other versions of GO. Just add such matrix to the workflow. I have added all versions 1.13 -> 1.17 but we can narrow that matrix to other versions.

MichalGuzieniuk:
@Eoghan1232 Can you please review it?

scripts/e2e_setup_cluster.sh
Eoghan1232: +1 on this.

test/e2e/custominjection_test.go
MichalGuzieniuk: Solved. I did all adaptations.

scripts/e2e_setup_cluster.sh
MichalGuzieniuk: I removed server_huge.yaml file from this change. But instead of using runtime feature, so controlSwitches, I stayed with the old solution. **injectHugepageDownApi** argument of webhook is added to the yaml by environment configuration script. I did it so, because I think that such approach requires less effort, and does not add additional dependency to test code.

test/e2e/e2e_tests_suite_test.go
```diff
@@ -73,3 +75,12 @@ var _ = BeforeSuite(func(done Done) {
 
 	close(done)
 }, 60)
+
+func hugepageOrSkip() {
```
Eoghan1232, line 79: Nit.
hugepageOrSkip -> HugePageOrSkip

To keep naming consistent with other functions in the file.

Eoghan1232 (approved):
lgtm

test/e2e/e2e_tests_suite_test.go
MichalGuzieniuk: From my point of view, the name of that function is correct. I don't want to export it outside e2e package and due to that it starts from lower case.

Eoghan1232: Oh yes, you're right.

SchSeba (approved):

Issue #117, opened 2021-10-19
Create a Deployment yaml for network-resource-injector


adrianchiris:
currently our Deployment example cosists of:
- permissions (auth.yaml): service account, roles, clusterRoles, roleBindings, clusterRoleBindings
- pod(server.yaml):
- service (service.yaml): k8s service exposing NRI
- webhook config (webhook.yaml): MutatingWebhookConfiguration

we should have a deployment instead of a pod so that there may be multiple backends to the service as well as have NRI pods lifecycle managed by a controller to handle placement and restarts.

while at it, its worth doing some cleanup in `auth.yaml` to remove certificate signing request as they are no longer used

PR #122, opened 2022-03-29
Use deployment for network-resource-injector and support pdb

VivekThrivikraman-est:
This changeset consists of following changes:

-  Use deployment instead of a pod 
- Support pdb
- Change in initcontainer code for supporting deployment: 
   The existing code would only have one initcontainer which creates certificate/key and webhook config with the generated 
   key. This would cause issues with deployments as multiple initcontainer's would race for creation of certificate/key and 
   webhook configuration. The new implementation uses secret to share certificate and key across all initcontainers and also 
   uses the secret resource as a lock also(ie to decide which initcontainer actually creates the certificate/key and webhook 
   config, avoiding a race).

Fixes: https://github.com/k8snetworkplumbingwg/network-resources-injector/issues/117

### DIFF
--- a/pkg/installer/installer.go
+++ b/pkg/installer/installer.go
@@ -16,9 +16,12 @@ package installer
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"strings"
+	"time"
 
 	"github.com/cloudflare/cfssl/csr"
 	"github.com/cloudflare/cfssl/helpers"
@@ -32,6 +35,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
@@ -40,6 +44,7 @@ var (
 	clientset kubernetes.Interface
 	namespace string
 	prefix    string
+	podName   string
 )
 
 const keyBitLength = 3072
@@ -232,6 +237,7 @@ func Install(k8sNamespace, namePrefix, failurePolicy string) {
 	if err != nil {
 		glog.Fatalf("error setting up Kubernetes client: %s", err)
 	}
+	populatePodName()
 
 	namespace = k8sNamespace
 	prefix = namePrefix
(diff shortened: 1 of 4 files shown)


SchSeba:
Hi @VivekThrivikraman-est is there a reason you can't use the https://github.com/k8snetworkplumbingwg/sriov-network-operator?

It will deploy the resource injector as part of the deployment.

SchSeba:
@zshi-redhat @adrianchiris do you think we should remove/document the is recommended to use the sriov operator on all the operand repos like device-plugin,cni and this one?

pkg/installer/installer.go
```diff
@@ -278,3 +297,84 @@ func Install(k8sNamespace, namePrefix, failurePolicy string) {
...
+}
+
+func populatePodName() {
+	var isPodNameAvailable bool
+	podName, isPodNameAvailable = os.LookupEnv("POD_NAME")
+	if !isPodNameAvailable {
+		glog.Fatal(errors.New("pod name not set as environment variable"))
+	}
+	glog.Info("Pod Name set:", podName)
+	// pod, err := clientset.CoreV1().Pods(namespace).Get(context.TODO(), leaseName, metav1.GetOptions{TypeMeta: metav1.TypeMeta{Kind: "Pod", APIVersion: "v1"}})
```
adrianchiris, line 355: nit: delete this

pkg/installer/installer.go
```diff
@@ -278,3 +297,84 @@ func Install(k8sNamespace, namePrefix, failurePolicy string) {
...
+
+}
+
+func waitForCertDetailsUpdation() error {
```
adrianchiris, line 359: nit: waitForCertDetailsUpdate

pkg/installer/installer.go
```diff
@@ -278,3 +297,84 @@ func Install(k8sNamespace, namePrefix, failurePolicy string) {
...
+func createSecret(ctx context.Context, certificate, key []byte, certFilename, keyFilename string) error {
+	ownerRef, err := getOwnerReference()
+	if err != nil {
+		glog.Fatalf("Failed fetching owner reference for the pod:%v", err)
+	}
+	// Set owner reference so that on deleting deployment the secret is also deleted,
+	// with this every new installation will create a new certificate and webhook config.
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            "network-resources-injector",
```
adrianchiris, line 310: can you put this name as constant ?

adrianchiris:
>Hi @VivekThrivikraman-est is there a reason you can't use the https://github.com/k8snetworkplumbingwg/sriov-network-operator?
>
> It will deploy the resource injector as part of the deployment

In sriov network operator we rely on externally provisioned cert and keys (either via openshift, cert manager or env variable)
it does not invoke the installer as init container.

this deployment provides a different way of doing it IIUC

pkg/installer/installer.go
VivekThrivikraman-est: Done.

pkg/installer/installer.go
VivekThrivikraman-est: Done.

pkg/installer/installer.go
VivekThrivikraman-est: Done.

adrianchiris (approved):

Eoghan1232 (approved):
Overall, code looks good to me.

zshi-redhat:
> @zshi-redhat @adrianchiris do you think we should remove/document the is recommended to use the sriov operator on all the operand repos like device-plugin,cni and this one?

yes, I think we can document the option to use sriov operator as a integrated solution for managing operands like plugin, cni or injector. There might be use cases where this injector is used for non-sriov case (I remember there was), so we would still support running standalone injector.

zshi-redhat:
I'll go ahead and merge this PR since we have two approvels.

Issue #124, opened 2022-06-27
Health check for webhook server


rthakur-est:
Support health check monitoring of NRI webhook server using k8s probes.

PR #125, opened 2022-06-27
health check for webhook server

rthakur-est:
This change adds support for health check endpoint for webhook.
Enabling of health check is made optional since kubelet doesn't support mTLS for probes, so it will only work for "insecure" mode. On enabling health check, client authentication is disabled.

#124 

Signed-off-by: Ravindra Thakur <ravindra.nath.thakur@est.tech>

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -51,6 +51,7 @@ func main() {
 	key := flag.String("tls-private-key-file", "key.pem", "File containing the default x509 private key matching --tls-cert-file.")
 	insecure := flag.Bool("insecure", false, "Disable adding client CA to server TLS endpoint --insecure")
 	flag.Var(&clientCAPaths, "client-ca", "File containing client CA. This flag is repeatable if more than one client CA needs to be added to server")
+	healthCheckPort := flag.Int("health-check-port", 8444, "The port to use for health check monitoring")
 
 	// do initialization of control switches flags
 	controlSwitches := controlswitches.SetupControlSwitchesFlags()
@@ -62,7 +63,7 @@ func main() {
 	controlSwitches.InitControlSwitches()
 	glog.Infof("controlSwitches: %+v", *controlSwitches)
 
-	if *port < 1024 || *port > 65535 {
+	if !isValidPort(*port) {
 		glog.Fatalf("invalid port number. Choose between 1024 and 65535")
 	}
 
@@ -80,6 +81,25 @@ func main() {
 
 	if namespace = os.Getenv("NAMESPACE"); namespace == "" {
 		namespace = "kube-system"
+	}
+
+	if !isValidPort(*healthCheckPort) {
+		glog.Fatalf("Invalid health check port number. Choose between 1024 and 65535")
+	} else if *healthCheckPort == *port {
+		glog.Fatalf("Health check port should be different from port")
+	} else {
+		go func() {
+			addr := fmt.Sprintf("%s:%d", *address, *healthCheckPort)
+			mux := http.NewServeMux()
+
+			mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+			})
+			err := http.ListenAndServe(addr, mux)
+			if err != nil {
+				glog.Fatalf("error starting health check server: %v", err)
+			}
+		}()
 	}
 
 	glog.Infof("starting mutating admission controller for network resources injection")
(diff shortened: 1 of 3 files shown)


cmd/webhook/main.go
```diff
@@ -81,6 +82,12 @@ func main() {
 	if namespace = os.Getenv("NAMESPACE"); namespace == "" {
 		namespace = "kube-system"
 	}
+	if *enableHealthCheck == true {
+		if *insecure == false {
```
adrianchiris, line 86: i think we should not implicitly set insecure to true, but rather fail and expect the user to set it WDYT ?

cmd/webhook/main.go
```diff
@@ -124,6 +131,10 @@ func main() {
 			webhook.MutateHandler(w, r)
 		})
 
+		http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
```
adrianchiris, line 134: can you add a deployment example with health check (with liveness probe) in https://github.com/k8snetworkplumbingwg/network-resources-injector/blob/master/deployments/server.yaml ?


as in k8s liveness http probe example:
```
apiVersion: v1

kind: Pod

metadata:

  labels:

    test: liveness

  name: liveness-http

spec:

  containers:

  - name: liveness

    image: k8s.gcr.io/liveness

    args:

    - /server

    livenessProbe:

      httpGet:

        path: /healthz

        port: 8080

        httpHeaders:

        - name: Custom-Header

          value: Awesome

      initialDelaySeconds: 3

      periodSeconds: 3

```

adrianchiris (changes requested):

cmd/webhook/main.go
rthakur-est: I am okay with this. Will make the change.

adrianchiris:
hi @rthakur-est been thinking about this more.

i think using mTLS is useful for mutating webhook. 

I think we should take a similar pattern as controller-runtime and operator-sdk with regards to mutating webhook and just create a separate http server to serve /healthz endpoint.

https://github.com/kubernetes-sigs/controller-runtime/blob/365ae09c4c6c466edaa91c919f7654944057e0b6/pkg/manager/internal.go#L340

note that in that implementation the healthcheck is non secure (regular HTTP)
for NRI we can either do the same or take the more secure path and use server side tls.

that way we dont need to give up on mTLS for the webhook if health check is enabled. WDYT ?

also as a note, whatever we end up with should be updated in README under : https://github.com/k8snetworkplumbingwg/network-resources-injector#additional-features

rthakur-est:
> hi @rthakur-est been thinking about this more.
> 
> i think using mTLS is useful for mutating webhook.
> 
> I think we should take a similar pattern as controller-runtime and operator-sdk with regards to mutating webhook and just create a separate http server to serve /healthz endpoint.
> 
> https://github.com/kubernetes-sigs/controller-runtime/blob/365ae09c4c6c466edaa91c919f7654944057e0b6/pkg/manager/internal.go#L340
> 
> note that in that implementation the healthcheck is non secure (regular HTTP) for NRI we can either do the same or take the more secure path and use server side tls.
> 
> that way we dont need to give up on mTLS for the webhook if health check is enabled. WDYT ?
> 
> also as a note, whatever we end up with should be updated in README under : https://github.com/k8snetworkplumbingwg/network-resources-injector#additional-features

@adrianchiris I think this way it would be better. End users won't need to lower down the security of webhook to monitor the health of the server. I will update the patch with this approach.

cmd/webhook/main.go
```diff
@@ -82,6 +83,23 @@ func main() {
 		namespace = "kube-system"
 	}
 
+	if *healthCheckPort < 1024 || *healthCheckPort > 65535 {
```
e0ne, line 86: If we want to don't allow 1-1023 ports usage it should be documented

e0ne (changes requested):

cmd/webhook/main.go
adrianchiris: these are regarded as "well known ports" and should not be used (reserved by IANA).

no need to document IMO

adrianchiris, line 86: nit: this and the check at L66 are the same, mind putting it into a helper func e.g `isPortValid(port) bool`

adrianchiris, line 86: you should check that healthCheckPort is not equal to port

e0ne (approved):
LGTM

adrianchiris (approved):

eoghanlawless (commented):
LGTM

rthakur-est:
can this PR be merged now?

adrianchiris:
Yes ! thanks work working on this

rthakur-est:
Thanks Adrian.

PR #136, opened 2023-05-17
Enable race detection in unit tests

AlinaSecret:
This pull request enables race detection in unit tests by modifying the go test command in test.sh. The test configuration is updated to include the '-race' flag, allowing thorough race detection during testing. 

Race detection can help identifying potential data race conditions, which can lead to hard-to-debug issues in concurrent code. By enabling race detection in our unit tests, we can detect and address these race conditions, ensuring the reliability and stability of our code. (for further reference view: https://go.dev/doc/articles/race_detector)


Eoghan1232 (approved):

adrianchiris:
we got some issues with E2E tests :\ 
i assume its because master label changed to control-plane

Eoghan1232:
> we got some issues with E2E tests :\ i assume its because master label changed to control-plane

most likely: 
```
error: taint "node-role.kubernetes.io/master:NoSchedule" not found
```

adrianchiris:
BTW, as an FYI, -race flag increases resource consumtion by X5-10 of the binary. so i wouldnt stick it everywhere.

if there are projects that are not single threaded/goroutine, AND the flows are susceptible to such issues, then I would add it.

adrianchiris (approved):

PR #140, opened 2023-06-30
Support pods without volumes

zeeke:
When a pod has `.spec.automountServiceAccountToken=false`, it has no volumes nor containers[*].volumeMounts, and the mutating webhook must produce a patch to create those arrays.

Sample error:

```
Internal error occurred: add operation does not apply: doc is missing path: "/spec/containers/0/volumeMounts/-": missing value

```


zeeke:
@Eoghan1232 , @SchSeba , @adrianchiris Please, take a look

Eoghan1232 (approved):
lgtm!

PR #144, opened 2023-10-23
Bump golang 1.21 to k8s 1.28.3

cgoncalves:


cgoncalves:
setup-go is installing golang version `1.2` instead of `1.20`. Need to quote versions.

adrianchiris (approved):

PR #147, opened 2024-01-09
Enable FIPS Compliance

vrindle:
CGO Flag needs to be 1 and ssl should be enabled in order to enable FIPS.


wizhaoredhat:
@SchSeba PTAL

SchSeba:
in general looks good but there is a problem in the CI

SchSeba:
closing his one we already fix the fips issues

PR #167, opened 2025-06-18
add support for any hugepage size on api

SchSeba:


adrianchiris (approved):
should there be an update to the README section ? "Expose Hugepages via Downward API" while at it ?

SchSeba:
you are right!

done

adrianchiris (approved):

SchSeba:
Thanks for the help!

Merging

PR #205, opened 2026-04-13
harden webhook TLS configuration flags

SchSeba:
Use component-base TLS parsing while enforcing a TLS 1.2 floor and rejecting insecure cipher suites so operator input cannot silently weaken webhook security. Add parsing tests and deployment docs to make the new TLS behavior explicit and safer to operate.


SchSeba:
@zeeke please take a look when you have time

pkg/webhook/tlsconfig.go
```diff
@@ -0,0 +1,70 @@
+package webhook
+
+import (
+	"crypto/tls"
+	"fmt"
+	"strings"
+
+	cliflag "k8s.io/component-base/cli/flag"
```
adrianchiris, line 8: the functions used from cliflag are basic wrapper around tls pkg.

WDYT about removing this dependency and just implementing the logic here ?

SchSeba: I prefer to leave this one so when core kubernetes add support for curves we just need to bump the package.
this way we are sure we follow what code kubernetes uses.

let me know what do you think

adrianchiris: after thinking about it a bit more, ack i think it makes sense.

adrianchiris (approved):

PR #216, opened 2026-06-11
add --tls-curve-preferences flag for webhook TLS hardening

SchSeba:
Allow administrators to control the TLS curve/group negotiation order via a new CLI flag, complementing the existing --tls-cipher-suites and --tls-min-version flags. Supported groups follow the IANA "TLS Supported Groups" registry: X25519, secp256r1, secp384r1, secp521r1, X25519MLKEM768.

## Summary by CodeRabbit

* **New Features**
  * Added a `--tls-curve-preferences` flag to configure allowed TLS curve preferences for the webhook server.
* **Deployment Changes**
  * Updated the server Kubernetes manifest to include an optional (commented) `tls-curve-preferences` setting.
* **Tests**
  * Added coverage for parsing TLS curve preferences, including valid formats, whitespace handling, and error cases.

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -57,6 +57,12 @@ func main() {
 		"Comma-separated list of TLS 1.2 and earlier cipher suites (for example TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256). Insecure cipher suites are rejected. If empty, uses Go runtime defaults.")
 	tlsMinVersion := flag.String("tls-min-version", "",
 		"Minimum TLS version (VersionTLS12, VersionTLS13). Values below VersionTLS12 are rejected. If empty, uses component-base default (VersionTLS12).")
+	tlsCurvePreferences := flag.String("tls-curve-preferences", "",
+		"Comma-separated list of numeric Go crypto/tls CurveID values, "+
+			"as the allowed key exchange mechanisms for the server. "+
+			"The supported values depend on the Go version used. "+
+			"See https://pkg.go.dev/crypto/tls#CurveID for values supported for each Go version. "+
+			"If empty, uses Go runtime defaults.")
 
 	// do initialization of control switches flags
 	controlSwitches := controlswitches.SetupControlSwitchesFlags()
@@ -129,7 +135,12 @@ func main() {
 		glog.Fatalf("invalid tls-min-version flag: %v", err)
 	}
 
-	glog.Infof("TLS config: minVersion=%#x, cipherSuites=%v", minVersion, cipherSuites)
+	curvePreferences, err := webhook.ParseCurvePreferences(*tlsCurvePreferences)
+	if err != nil {
+		glog.Fatalf("Error parsing TLS curve preferences: %v", err)
+	}
+
+	glog.Infof("TLS config: minVersion=%#x, cipherSuites=%v, curvePreferences=%v", minVersion, cipherSuites, curvePreferences)
 
 	/* init API client */
 	clientset := webhook.SetupInClusterClient()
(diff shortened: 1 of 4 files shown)


pkg/webhook/tlsconfig.go
```diff
@@ -51,6 +51,45 @@ func ParseTLSCipherSuites(cipherSuites string) ([]uint16, error) {
...
 
+// supportedCurves maps TLS group/curve name strings to their corresponding tls.CurveID.
+// Names follow IANA's "TLS Supported Groups" registry.
+var supportedCurves = map[string]tls.CurveID{
+	"X25519":         tls.X25519,
+	"secp256r1":      tls.CurveP256,
+	"secp384r1":      tls.CurveP384,
+	"secp521r1":      tls.CurveP521,
+	"X25519MLKEM768": tls.X25519MLKEM768,
+}
```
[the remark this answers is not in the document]
SchSeba: we don't need this the project should compile on 1.25+

pkg/webhook/tlsconfig_test.go
```diff
@@ -152,6 +152,72 @@ var _ = Describe("TLS Configuration", func() {
...
+				curves, err := ParseCurvePreferences(name)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(curves).To(HaveLen(1))
+				Expect(curves[0]).To(Equal(expected))
+			},
+			Entry("X25519", "X25519", tls.X25519),
+			Entry("secp256r1", "secp256r1", tls.CurveP256),
+			Entry("secp384r1", "secp384r1", tls.CurveP384),
+			Entry("secp521r1", "secp521r1", tls.CurveP521),
+			Entry("X25519MLKEM768", "X25519MLKEM768", tls.X25519MLKEM768),
```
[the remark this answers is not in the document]
SchSeba: we don't need this the project should compile on 1.25+

SchSeba:
based on https://github.com/openshift/api/pull/2583 we also need to support the curves.
that is what this PR adds to the project

cmd/webhook/main.go
```diff
@@ -57,6 +57,8 @@ func main() {
 		"Comma-separated list of TLS 1.2 and earlier cipher suites (for example TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256). Insecure cipher suites are rejected. If empty, uses Go runtime defaults.")
 	tlsMinVersion := flag.String("tls-min-version", "",
 		"Minimum TLS version (VersionTLS12, VersionTLS13). Values below VersionTLS12 are rejected. If empty, uses component-base default (VersionTLS12).")
+	tlsCurvePreferences := flag.String("tls-curve-preferences", "",
```
adrianchiris, line 60: can we name it: `tls-supported-groups` ?

from an A.I overview it seems more suitable?


=====
What “supported groups” means
In TLS 1.2 and 1.3, the client can send an extension called supported_groups (originally elliptic_curves), which is just an ordered list of group IDs it is willing to use for key agreement. The server picks one group from this list (that it also supports) and uses it for the Diffie–Hellman key exchange that creates the shared secret.

A “group” here is a specific mathematical object suitable for Diffie–Hellman, such as an elliptic curve over a prime field (for ECDHE) or a finite field group for classic DHE. These are centrally assigned and documented by IANA in the “TLS Supported Groups” registry so implementations can interoperate.

adrianchiris: if you agree with this, we should rename functions/vars/structs in this PR accordingly

adrianchiris: note that indeed this extention was named eliptic_curves but was renamed.

SchSeba: HI prefer to maintain compatibility with kubernetes if possible https://github.com/kubernetes/component-base/commit/e98593989947d36e0b32448b314252ab3c3fc3dc

adrianchiris: ack. makes sense, thx for the ref.

i see they gave up on string mapping and just forward the user to go crypto consts.

see:
https://github.com/kubernetes/kubernetes/blob/8b4b9515d8c137e880be16d6070e7f4ffeebee83/staging/src/k8s.io/apiserver/pkg/server/options/serving.go#L200

should we do the same ?

pkg/webhook/tlsconfig.go
```diff
@@ -51,6 +52,35 @@
...
+	for _, part := range parts {
+		s := strings.TrimSpace(part)
+		if s == "" {
+			return nil, fmt.Errorf("empty TLS CurveID value")
+		}
+		v, err := strconv.ParseInt(s, 10, 32)
+		if err != nil {
+			return nil, fmt.Errorf("invalid TLS CurveID value %q: %w", s, err)
+		}
+		curves = append(curves, tls.CurveID(v))
```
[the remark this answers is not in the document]
adrianchiris: curveID is uint16 maybe we need to tighten up the parsing

SchSeba:
Hi @adrianchiris let me know if we merge this one

adrianchiris (approved):
i thought i approved this one already. sorry.

LGTM !

PR #224, opened 2026-06-29
webhook: Skip duplicate podnetinfo volume and mount injection

nirdothan:
Multiple webhook instances may run concurrently, causing the webhook to fire multiple times for the same pod. Resource injection already handles this by checking for existing resources before adding them. Volume injection lacked this protection, causing pod creation failures when podnetinfo volume or mounts were injected more than once [1].

Add duplicate-detection logic to volume and volume mount injection.

[1] https://redhat.atlassian.net/browse/CNV-91360

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents duplicate pod volume and volume-mount injections, making webhook updates idempotent.
  * Keeps existing pod configuration unchanged when the target volume or mount is already present.
* **Tests**
  * Added coverage for duplicate-prevention behavior to confirm patches are only applied when needed.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"net/http"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


zeeke (approved):
LGTM

SchSeba (approved):
LGTM

PR #231, opened 2026-08-03
Validate network selection fields and reject cross-namespace NAD refrences

yahlifried:
Move character validation (lowercase alphanumeric + hyphens) from the CSV-only parsePodNetworkSelectionElement into parsePodNetworkSelections so it covers both JSON and CSV annotation formats. Add nil-element guard for JSON arrays. Reject NAD references where the specified namespace differs from the pod's own namespace.

_Cross-namespace NAD references are now rejected unconditionally. This is a breaking change for users who reference NADs from other namespaces - if needed, a --namespace-isolation flag (mirroring Multus's pattern) can be added in a follow-up to make this configurable._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for network selection names, namespaces, and interfaces.
  * Rejected empty values, null entries, missing fields, invalid characters, path traversal patterns, and overlong values.
  * Improved handling of invalid or empty network selection input.
  * Prevented duplicate webhook volume and volume-mount injection when equivalent entries already exist.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -61,6 +61,7 @@ const (
 
 var (
 	HugepageRegex         = regexp.MustCompile(`^hugepages-(.+)$`)
+	validNameRegex        = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`)
 	clientset             kubernetes.Interface
 	nadCache              netcache.NetAttachDefCacheService
 	userDefinedInjections *userdefinedinjections.UserDefinedInjections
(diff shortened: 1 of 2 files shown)


pkg/webhook/webhook.go
```diff
@@ -271,7 +272,26 @@ func parsePodNetworkSelections(podNetworks, defaultNamespace string) ([]*multus.
...
+			err := errors.New("invalid network selection: null element in list")
+			glog.Error(err)
+			return nil, err
+		}
+		for _, pair := range []struct{ label, value string }{
+			{"name", networkSelection.Name},
+			{"namespace", networkSelection.Namespace},
+			{"interface", networkSelection.InterfaceRequest},
+		} {
+			if pair.value != "" && !validNameRegex.MatchString(pair.value) {
```
SchSeba: lets add `maximum length enforced on validated fields`

pkg/webhook/webhook.go
```diff
@@ -271,7 +272,26 @@ func parsePodNetworkSelections(podNetworks, defaultNamespace string) ([]*multus.
 		}
 	}
 
-	/* fill missing namespaces with default value */
+	for _, networkSelection := range networkSelections {
```
SchSeba, line 277: [HIGH] Empty name field bypasses validation and hits API list endpoint
webhook.go:281-291
Category: security

The validation loop skips empty values (if pair.value != "" && ...). A JSON annotation like [{"namespace": "default"}] or [{"name": "", "namespace": "default"}] produces a NetworkSelectionElement with an empty Name that passes validation. This empty name is then interpolated into the API path at line 363:


/apis/k8s.cni.cncf.io/v1/namespaces/default/network-attachment-definitions/

pkg/webhook/webhook.go
```diff
@@ -286,6 +306,12 @@ func parsePodNetworkSelections(podNetworks, defaultNamespace string) ([]*multus.
 			} else {
 				networkSelection.Namespace = defaultNamespace
 			}
+		} else if networkSelection.Namespace != defaultNamespace {
```
SchSeba: this is wrong, if I remember right in multus this is approved that is why namespace exist in the annotation. it will break a supported feature

pkg/webhook/webhook.go
yahlifried: Done

pkg/webhook/webhook.go
yahlifried: Done

pkg/webhook/webhook.go
yahlifried: Done

SchSeba:
Hi @yahlifried please rebase this one

PR #233, opened 2026-08-04
Add /readyz endpoint and readinessProbe to network-resources-injector

schsebabot:
This PR adds a `/readyz` readiness endpoint to the network-resources-injector webhook server and introduces a corresponding `readinessProbe` in the Kubernetes deployment manifest. It also enhances the existing `livenessProbe` configuration with additional robustness fields and adds unit tests to validate both health check endpoints.

## Key Changes

- **`cmd/webhook/main.go`**: Added a `/readyz` HTTP handler alongside the existing `/healthz` handler in the health check server goroutine. Both endpoints return `HTTP 200 OK`.

- **`deployments/server.yaml`**: Added a `readinessProbe` targeting the new `/readyz` endpoint on port `8444` with `initialDelaySeconds: 5` and `periodSeconds: 10`. Enhanced the existing `livenessProbe` with explicit fields: `failureThreshold`, `scheme`, `successThreshold`, and `timeoutSeconds` for improved clarity and robustness.

- **Unit tests**: Added tests to verify that both `/healthz` and `/readyz` endpoints respond with `HTTP 200 OK`, ensuring correctness of the health check server behavior.

## Notable Implementation Decisions

- The `/readyz` handler uses the same simple `200 OK` response as `/healthz`, keeping the implementation straightforward. Both probes share port `8444` (the existing health check server port), avoiding the need to expose an additional port.
- The `livenessProbe` was enhanced with explicit default values (`scheme: HTTP`, `failureThreshold: 3`, etc.) to make the configuration self-documenting and consistent with Kubernetes best practices.

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -106,6 +106,9 @@ func main() {
 			mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
 				w.WriteHeader(http.StatusOK)
 			})
+			mux.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+			})
 			err := http.ListenAndServe(addr, mux)
 			if err != nil {
 				glog.Fatalf("error starting health check server: %v", err)
--- a/deployments/server.yaml
+++ b/deployments/server.yaml
@@ -72,11 +72,21 @@ spec:
             memory: "200Mi"
             cpu: "500m"
         livenessProbe:
+          failureThreshold: 3
           httpGet:
             path: /healthz
             port: 8444
-          initialDelaySeconds: 10
-          periodSeconds: 5
+            scheme: HTTP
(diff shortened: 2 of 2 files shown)


cmd/webhook/main_test.go
```diff
@@ -0,0 +1,39 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestHealthEndpoints(t *testing.T) {
```
SchSeba: this test doesn't really run the resouce-injector code so it's a bit point less.
lets remove this file and skip the test for this

schsebabot: Good point — the test was only exercising the HTTP handler in isolation without any of the actual injector logic, so it doesn't add much value. I've removed `cmd/webhook/main_test.go` in the latest push. Thanks for the feedback!

zeeke (approved):

SchSeba (approved):

PR #235, opened 2026-08-10
add the option to pass --health-check-port=8444 as a parameter

schsebabot:
This PR adds the ability to configure the health check port via a `--health-check-port` command-line parameter, complementing the existing health check functionality added in [#233](https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/233). Previously, the health check server port was hardcoded with no way to override it at runtime.

## Key Changes

- Added `--health-check-port` flag to the binary's argument parsing, allowing users to specify a custom port (e.g., `--health-check-port=8444`)
- Wired the new port parameter through to the health check server initialization so the configured value is actually used
- Updated the Helm chart / Kubernetes deployment manifests to expose the `healthCheckPort` as a configurable value
- Added the corresponding `containerPort` entry in the pod spec so the health check port is properly declared and accessible within the cluster

## Notable Implementation Decisions

- The flag is optional with a sensible default so existing deployments are not affected
- The port is propagated through both the CLI argument and the container port definition to ensure full end-to-end configurability without requiring manual manifest edits

## Summary by CodeRabbit

* **New Features**
  * Added dedicated health and readiness endpoints for webhook deployments.
  * Exposed separate webhook and health-check ports for service traffic.

* **Bug Fixes**
  * Improved liveness and readiness probe configuration with explicit ports, HTTP schemes, success thresholds, and timeouts.
  * Ensured health probes return successful responses on the main webhook port.

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -99,6 +99,7 @@ func main() {
 	} else if *healthCheckPort == *port {
 		glog.Fatalf("Health check port should be different from port")
 	} else {
+		glog.Infof("starting health check server on %s:%d", *address, *healthCheckPort)
 		go func() {
 			addr := fmt.Sprintf("%s:%d", *address, *healthCheckPort)
 			mux := http.NewServeMux()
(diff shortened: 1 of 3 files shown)


SchSeba:
closing this one

